Issue #181, opened 2024-11-26
CI/CD for Iceberg/Hadoop benchmarking is broken after publishing build scripts changed


oleg-lvovitch-aws:
### Analytics Accelerator Library for Amazon S3 Version

0.0.1

### AWS Region

N/A

### Describe the running environment

GitHub Ci/CD

### What happened?

After we changed the publishing and shading, the CI/CD scripts are broken at the time of publishing into the local Maven:`Failed to execute goal org.apache.maven.plugins:maven-install-plugin:2.4:install-file (default-cli) on project standalone-pom: The specified file '/home/runner/work/analytics-accelerator-s3/analytics-accelerator-s3/input-stream/build/libs/input-stream-all.jar'` - see [Build Hadoop Artifacts](https://github.com/awslabs/analytics-accelerator-s3/actions/runs/12033795225/job/33548790589)

`Error:  Failed to execute goal org.apache.maven.plugins:maven-install-plugin:2.4:install-file (default-cli) on project standalone-pom: The specified file '/home/runner/work/analytics-accelerator-s3/analytics-accelerator-s3/input-stream/build/libs/input-stream.jar' not exists` - see [Build Iceberg Artifacts](https://github.com/awslabs/analytics-accelerator-s3/actions/runs/12033795225/job/33548791330)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

CsengerG:
CICD is now green.  @oleg-lvovitch-aws, can we close or did you want to drive any other improvements through this?

oleg-lvovitch-aws:
Thank you for closing on this!

PR #182, opened 2024-11-26
Remove unnecessary Maven publish step

CsengerG:
This commit removes a stanza from the workflow that was copying old artefacts (common, input-stream, object-client) into Maven, explicitly via Maven commands.

This should not be needed because we have `publishToMavenLocal` and also we do not generate or publish these artefacts anymore.

#### Relevant issues

https://github.com/awslabs/analytics-accelerator-s3/issues/181

No

No

It will be tested upon push, I have no way to run GitHub automation locally.

---

> By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and I agree to the terms of the [Developer Certificate of Origin (DCO)](https://developercertificate.org/).

### DIFF
--- a/.github/workflows/build-upload.yml
+++ b/.github/workflows/build-upload.yml
@@ -120,12 +120,6 @@ jobs:
         run: |
           ./gradlew publishToMavenLocal
 
-      - name: TEMPORARY publish updated versions to local Maven
-        run: |
-          mvn install:install-file -Dfile=input-stream/build/libs/input-stream.jar -DgroupId=com.amazon.connector.s3 -DartifactId=input-stream -Dversion=0.0.1 -Dpackaging=jar -DgeneratePom=true
-          mvn install:install-file -Dfile=object-client/build/libs/object-client.jar -DgroupId=com.amazon.connector.s3 -DartifactId=object-client -Dversion=0.0.1 -Dpackaging=jar -DgeneratePom=true
-          mvn install:install-file -Dfile=common/build/libs/common.jar -DgroupId=com.amazon.connector.s3 -DartifactId=common -Dversion=0.0.1 -Dpackaging=jar -DgeneratePom=true
-
       - name: Setup Iceberg SSH deploy key
         uses: webfactory/ssh-agent@v0.9.0
         with:
--- a/.github/workflows/build-upload.yml
+++ b/.github/workflows/build-upload.yml
@@ -21,8 +21,8 @@ permissions:
       contents: read    # This is required for actions/checkout
 
 jobs:
-  BuildCfS3AndUploadArtifact:
-    name: Build CFS3 artifacts
+  BuildAnalyticsAcceleratorAndUploadArtifact:
+    name: Build library artifacts
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/build-upload.yml
```diff
@@ -120,12 +120,6 @@ jobs:
         run: |
           ./gradlew publishToMavenLocal
```
oleg-lvovitch-aws, line 122: So the idea here is that `publishToMavenLocal` will push what we have just built to the local `.m2`, correct? One thing we should probably test is that the Hadoop/Iceberg build will pick up the latest version we have just pushed, as opposed to the one that is available in the remote repo, otherwise we will be benchmarking against the latest release over the most recent codebase.

One way to make this true for sure is to push the local maven under a different, synthetic version (e.g. `0.0.1-SNAPSHOT`) to guarantee this. I am open to whatever works.

CsengerG: Added this synthetic version now, confirmed locally that it gets generated. It's a bit of an effort to make the whole pipeline use this `SNAPSHOT` version and possibly out of scope for this PR. I would prefer to merge this, see that it doesn't get stuck on what it failed on before and then wire everything through with this new JAR in a separate one.

CsengerG: Here is the new idea. This gives us a `./gradlew publishSnapshotPublicationToMavenLocal` task. I think as long as we use this and then the SNAPSHOT version everywhere, we should be safe of the risk you mention.  Does this sound right?

input-stream/build.gradle.kts
```diff
@@ -307,9 +306,43 @@ publishing {
                 }
             }
         }
+
+        create<MavenPublication>("snapshot") {
```
fuatbasik, line 310: wouldn't this publish but version, and version-snapshot to Maven?

CsengerG: I think `publish` publishes all publications (does it publish it to all repos?), whereas if you do `publishMavenJavaPublicationToSonatypeRepository` only, then this one gets skipped.

Which task do we use right now to publish to Maven? Do we just do a simple `publish`?

CsengerG: (ignore this, I've found a better way)

.github/workflows/build-upload.yml
CsengerG: (ignore this, I've found a better way via Gradle properties)

input-stream/build.gradle.kts
```diff
@@ -275,14 +275,15 @@ val signingEnabled = project.hasProperty("signingEnabled") && project.property("
...
             artifact(tasks["shadowJar"])
             artifact(tasks.named("customSourcesJar"))
             artifact(tasks.named("customJavadocJar"))
 
             groupId = group
             artifactId = artefact
-            version = currentVersion
+
+            val isSnapshot = findProperty("snapshotBuild") == "true"
+            version = if (isSnapshot) "SNAPSHOT" else currentVersion;
```
fuatbasik, line 286: should we set the version or classifier?

fuatbasik (approved):
approving with an open question

PR #188, opened 2024-12-03
Split footer requests into two

ahmarsuhail:
## Description of change

Previously, we were making a single 1MB request for the tail of the file. 

This 1MB was chosen as it was big enough to contain the fileMetadata + pageIndex structures for most file sizes. 

However for larger files, this 1MB is not enough. Further, the fileMetadata is ALWAYS required first and is much smaller than the the pageIndex data. To help with this, this PR:

* Splits the request into two. One smaller prefetch request for the fileMetadata only, and another larger prefetch request for the pageIndex structures.
* Introduces new configs that can be used to set prefetch request sizes for these file sizes.
* Renames some existing configs to better suit their purposed (footerCachingEnabled becomes footerPrefetchEnabled). 
* Adds config for pageIndex prefetching, as this data is not always required, depending on the engine. 

#### Relevant issues

N/A

It renames footerCachingEnabled to footerPrefetchEnabled, to better suit it's purpose. I do not believe this has any impact currently. 

N/A

Ran unit tests, and with in proc Spark application to test requests are split correctly.

---

### DIFF
--- a/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/LogicalIOConfiguration.java
+++ b/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/LogicalIOConfiguration.java
@@ -15,6 +15,8 @@
  */
 package software.amazon.s3.analyticsaccelerator.io.logical;
 
+import static software.amazon.s3.analyticsaccelerator.util.Constants.ONE_GB;
+import static software.amazon.s3.analyticsaccelerator.util.Constants.ONE_KB;
 import static software.amazon.s3.analyticsaccelerator.util.Constants.ONE_MB;
 
 import lombok.Builder;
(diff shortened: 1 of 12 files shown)


input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/LogicalIOConfiguration.java
```diff
@@ -28,22 +30,51 @@
...
+  private static final String PAGE_INDEX_PREFETCH_ENABLED_KEY = "page.index.prefetch.enabled";
+
+  @Builder.Default private long fileMetadataPrefetchSize = DEFAULT_FILE_METADATA_PREFETCH_SIZE;
+
+  private static final String FILE_METADATA_PREFETCH_SIZE_KEY = "file.metadata.prefetch.size";
+
+  @Builder.Default
+  private long largeFileMetadataPrefetchSize = DEFAULT_LARGE_FILE_METADATA_PREFETCH_SIZE;
+
+  private static final String LARGE_FILE_METADATA_PREFETCH_SIZE_KEY =
```
CsengerG, line 62: This is a total nit, but what do you think of reversing the hierarchy here?

I feel like `largeFileSize` is a property of `prefetch` (would suggest something like `prefetch.largeFileSize`), whereas `prefetch` is not a property of `large`.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/parquet/FooterPrefetchSize.java
```diff
@@ -0,0 +1,34 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package software.amazon.s3.analyticsaccelerator.io.logical.parquet;
+
+import lombok.Data;
+
+/** Stores sizes to prefetch for file metadata and page index structures. */
+@Data
```
CsengerG, line 21: Are we mutating this structure anywhere?  If not, would prefer an immutable structure (i.e., `@Value`).

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/util/Constants.java
```diff
@@ -22,6 +22,7 @@ private Constants() {}
 
   public static final int ONE_KB = 1024;
   public static final int ONE_MB = 1024 * 1024;
+  public static final int ONE_GB = 1024 * 1024 * 1024;
```
CsengerG, line 25: Should we make this `long`?  Someone doing simple maths on this -- even just multiplying it by 5 -- might overflow already.

CsengerG (approved):
Thanks! This is looking good, had some minor observations.

Issue #173, opened 2024-11-25
TPC-DS Q44 regression for the " Mixed size objects" dataset


oleg-lvovitch-aws:
### Analytics Accelerator Library for Amazon S3 Version

0.0.1

### AWS Region

`us-east-1`, though it applies broadly. 

### Describe the running environment

TPC DS benchmarking 

### What happened?

Q44 shows 8% regression by comparison with the baseline. 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ahmarsuhail:
This has now been resolved with https://github.com/awslabs/analytics-accelerator-s3/pull/189, and released with v0.0.2 https://github.com/awslabs/analytics-accelerator-s3/releases/tag/v0.0.2

PR #189, opened 2024-12-03
Prefetch dictionaries and column data separately

ahmarsuhail:
## Description of change

Highly selective queries, may only want to read the dictionaries, and only when the predicate matches, prefetch columns.

Previously, we were prefetching entire column chunks which was leading us to prefetch more data required. This PR tracks two different lists

1/ Recently read dictionaries
2/ Recently read columns

For queries like `select a, b, c from ss where c = 124`, we would always want to read the dictionary for c, but may not need column data for a, b and c, if the predicate does not match. With this change, we only prefetch column data a,b,c when reading for columns actually starts. 

That is, when the query engine reads the first column, say a, we will prefetch b and c. 

#### Relevant issues

See https://github.com/awslabs/analytics-accelerator-s3/issues/173 for more details

No

No

Tested with local Spark runs against a TPC-DS dataset. 

---

### DIFF
--- a/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/impl/ParquetColumnPrefetchStore.java
+++ b/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/impl/ParquetColumnPrefetchStore.java
@@ -87,8 +87,13 @@ public class ParquetColumnPrefetchStore {
    */
   private final Map<Integer, LinkedList<String>> recentlyReadColumnsPerSchema;
 
-
-
+  /**
+   * This is a mapping of schema and the recently read column dictionaries for it. This differs from
+   * the recentlyReadColumnsPerSchema, as it tracks the dictionaries being read only. These two
+   * lists can differ. For example, say there is a query "select ss_a, ss_b from store_sales where
+   * ss_c = 123". Then, for predicate matching, the dictionary for ss_c will be read, and column
+   * data will be required for ss_a and ss_b.
+   */
   private final Map<Integer, LinkedList<String>> recentlyReadDictionariesPerSchema;
 
   /**
@@ -109,6 +114,10 @@ public class ParquetColumnPrefetchStore {
    */
   private final Map<S3URI, List<Integer>> columnRowGroupsPrefetched;
 
+  /**
+   * Similar to columnRowGroupsPrefetched, this is a mapping of S3 URI's of Parquet files to a list
+   * of row group indexes prefetched for dictionaries only.
+   */
   private final Map<S3URI, List<Integer>> dictionaryRowGroupsPrefetched;
 
   private final LogicalIOConfiguration configuration;
@@ -160,8 +169,11 @@ public class ParquetColumnPrefetchStore {
    * @param configuration LogicalIO configuration
    * @param columnMappersStore Store of column mappings
    * @param recentlyReadColumnsPerSchema List of recent read columns for each schema
-   * @param columnRowGroupsPrefetched Map of Parquet file URI to row groups that have been prefetched for
-   *     it
+   * @param recentlyReadDictionariesPerSchema List of recent read dictionaries for each schema
+   * @param columnRowGroupsPrefetched Map of Parquet file URI to row groups that have been
+   *     prefetched for it
+   * @param dictionaryRowGroupsPrefetched Map of Parquet file URI to row groups that have been
+   *     prefetched for it
    */
   ParquetColumnPrefetchStore(
       LogicalIOConfiguration configuration,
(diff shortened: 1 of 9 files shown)


input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/impl/ParquetColumnPrefetchStore.java
```diff
@@ -103,7 +112,13 @@ public class ParquetColumnPrefetchStore {
    * will then be added to this rowGroupsPrefetched map, so that if another read happens to a column
    * in this row group, prefetches for the row group are not triggerred again.
    */
-  private final Map<S3URI, List<Integer>> rowGroupsPrefetched;
+  private final Map<S3URI, List<Integer>> columnRowGroupsPrefetched;
```
CsengerG, line 115: Not having full context on everything this confuses me slightly.  I think I understood `rowGroupsPrefetched` and what it stands for.  How does `columnRowGroupsPrefetched` differ now from the original use of `rowGroupsPrefetched`?

CsengerG: Maybe I get it now after looking at `dictionaryRowGroupsPrefetched` -- does `columnRowGroupsPrefetched` mean "the columns (i.e., data) within the row group prefetched" and `dictionaryRowGroupsPrefetched` is "the prefetched dictionaries within the row group"?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/parquet/ParquetPredictivePrefetchingTask.java
```diff
@@ -116,16 +116,27 @@ public ParquetPredictivePrefetchingTask(
...
-        // prefetched already.
-        prefetchCurrentRowGroup(columnMappers, columnMetadata);
 
-        List<ColumnMetadata> addedColumns =
-            addAdjacentColumnsInLength(columnMetadata, columnMappers, position, len);
-        addedColumns.add(columnMetadata);
+        // If the column has a dictionary and the length of the read is <= the size of the
+        // dictionary, then assume current read is for a dictionary only.
+        if (columnMetadata.getDictionaryOffset() != 0
+            && len <= (columnMetadata.getDataPageOffset() - columnMetadata.getDictionaryOffset())) {
```
CsengerG, line 127: (I know we have an item for refactoring.)

I wonder if this factors out nicely into a boolean helper method to make it easier to read the code (looking at these big if-else structures can be challenging sometimes) -- it might not, just asking.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/parquet/ParquetPredictivePrefetchingTask.java
```diff
@@ -184,21 +209,43 @@ public IOPlanExecution prefetchRecentColumns(
...
+                          recentColumn,
+                          this.s3Uri.getKey(),
+                          columnMetadata.getRowGroupIndex());
+                    } else {
+                      prefetchRanges.add(
+                          new Range(
+                              columnMetadata.getStartPos(),
+                              columnMetadata.getStartPos()
+                                  + columnMetadata.getCompressedSize()
+                                  - 1));
```
CsengerG, line 242: I am a bit vary of some of the maths around here (testability, readability and code duplication arguments).  One observation is that there is a lot of

```
new Range(
                              columnMetadata.x(),
                              columnMetadata.y()
                                  + (columnMetadata.x()
                                      - columnMetadata.y()
                                      - 1)));
```

style of code here.  Should this go into columnMetadata instead?  Might be simpler to read `prefetchRanges.add(columnMetadata.getRangeForDictionary())`.

CsengerG (approved):
Good stuff, thanks!  Overall, the logic looks good, I got myself confused about one or two things, I left some notes around that if we want to improve it now or in the future is your call.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/impl/ParquetColumnPrefetchStore.java
ahmarsuhail: yes, exactly

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/parquet/ParquetPredictivePrefetchingTask.java
ahmarsuhail: agreed, but this will need to be done somewhere. So this can either be done when constructing the metadata, in the ParquetMetadataParsingTask, or here. I'm ok with keeping it here for now, but can discuss how to improve this.

Issue #83, opened 2024-08-21
`S3SeekableInputStream.seek` doesn't allow seeking past the end of the stream


oleg-lvovitch-aws:
This is not the case for [S3A ](https://github.com/apache/hadoop/blob/f6c45e0bcf4aeaba31515e548dcc98b33245fe0e/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AInputStream.java#L325) and [S3FileIO](https://github.com/apache/iceberg/blob/f1076494c807447484b5a154b242611e98e3b582/aws/src/main/java/org/apache/iceberg/aws/s3/S3InputStream.java#L83). 

We should be able to support the same behavior

fuatbasik:
Release 0.0.2 includes this change. Closing.

PR #192, opened 2024-12-09
Add support to seek beyond end of stream

fuatbasik:
## Description of change
This pull request adds support to seek beyond stream end. 
Since seek is done lazily this is the correct behaviour. It also adds closed checks to read operations.

With this PR, we are reconciling S3SeekableStream behaviour with its' documentation. 

#### Relevant issues
[Issue#83 ](https://github.com/awslabs/analytics-accelerator-s3/issues/83)

Yes. Seek will no longer Throw EOF exception when position to seek is > contentLength.

Yes, Seek to beyond stream end will no longer throw. Read/Seek operations on closed stream will throw. 

Added new unit tests and property based tests. Removed %size from `seekChangesPosition` property tests to ensure validSizes can go beyond stream size. 

---

### DIFF
--- a/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStream.java
+++ b/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStream.java
@@ -15,7 +15,6 @@
  */
 package software.amazon.s3.analyticsaccelerator;
 
-import java.io.EOFException;
 import java.io.IOException;
 import lombok.NonNull;
 import software.amazon.s3.analyticsaccelerator.common.Preconditions;
@@ -38,6 +37,7 @@ public class S3SeekableInputStream extends SeekableInputStream {
   private final Telemetry telemetry;
   private final S3URI s3URI;
   private long position;
+  private boolean closed;
   private static final int EOF = -1;
 
   private static final String OPERATION_READ = "stream.read";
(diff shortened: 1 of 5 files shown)


input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStream.java
```diff
@@ -38,6 +37,7 @@ public class S3SeekableInputStream extends SeekableInputStream {
   private final Telemetry telemetry;
   private final S3URI s3URI;
   private long position;
+  private boolean closed;
```
CsengerG, line 40: Should this be Atomic?  What happens when two threads close at the same time?

fuatbasik: yes, it should be. But, we are already calling out Stream is not threadsafe see docs: `Don't share between threads. The current implementation is not thread safe` so I didn't invest time in making it threadsafe. (The current problem is with the position though, not sure adding 1 more variable makes sense. )

CsengerG: Ack, was just asking -- not very opinionated. Maybe an Atomic position is a middle ground here.

CsengerG (approved):
LGTM

PR #193, opened 2024-12-10
Changes for S3a integration to support s3 client

rajdchak:
## Description of change
Changes for S3a integration to support s3 client

Tested using integration tests.

---

### DIFF
--- a/object-client/src/main/java/software/amazon/s3/analyticsaccelerator/S3SdkObjectClient.java
+++ b/object-client/src/main/java/software/amazon/s3/analyticsaccelerator/S3SdkObjectClient.java
@@ -18,11 +18,16 @@ package software.amazon.s3.analyticsaccelerator;
 import java.util.concurrent.CompletableFuture;
 import lombok.Getter;
 import lombok.NonNull;
+import software.amazon.awssdk.awscore.AwsClient;
 import software.amazon.awssdk.awscore.AwsRequestOverrideConfiguration;
+import software.amazon.awssdk.core.ResponseInputStream;
 import software.amazon.awssdk.core.async.AsyncResponseTransformer;
 import software.amazon.awssdk.services.s3.S3AsyncClient;
+import software.amazon.awssdk.services.s3.S3Client;
 import software.amazon.awssdk.services.s3.model.GetObjectRequest;
+import software.amazon.awssdk.services.s3.model.GetObjectResponse;
 import software.amazon.awssdk.services.s3.model.HeadObjectRequest;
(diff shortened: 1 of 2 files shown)


object-client/src/main/java/software/amazon/s3/analyticsaccelerator/S3SdkObjectClient.java
```diff
@@ -33,59 +38,59 @@ public class S3SdkObjectClient implements ObjectClient {
   private static final String HEADER_USER_AGENT = "User-Agent";
   private static final String HEADER_REFERER = "Referer";
 
-  @Getter @NonNull private final S3AsyncClient s3AsyncClient;
+  @Getter @NonNull private final AwsClient s3Client;
```
CsengerG, line 41: How general of a class is `AwsClient`?  With this, are we opening ourselves up to accepting clients that are not S3 clients (e.g., is DynamoDBClient also an AwsClient)?  Do we need any validations around allowed implementations?

object-client/src/main/java/software/amazon/s3/analyticsaccelerator/S3SdkObjectClient.java
```diff
@@ -125,15 +130,24 @@ public CompletableFuture<ObjectMetadata> headObject(HeadRequest headRequest) {
...
                         .build()));
   }
 
+  private CompletableFuture<HeadObjectResponse> headObjectForClient(HeadObjectRequest request) {
+    if (s3Client instanceof S3AsyncClient) {
+      return ((S3AsyncClient) s3Client).headObject(request);
+    } else if (s3Client instanceof S3Client) {
+      return CompletableFuture.completedFuture(((S3Client) s3Client).headObject(request));
+    } else {
+      throw new UnsupportedOperationException("Unsupported client type for headObject operation");
```
CsengerG, line 147: I think this is what my previous comment is touching on.  This validation is a bit late in my mind, we should not accept client implementations that cannot perform an S3 HeadObject.

CsengerG (commented):
Thanks! Left some comments for discussion.

object-client/src/main/java/software/amazon/s3/analyticsaccelerator/S3SdkObjectClient.java
fuatbasik: wait a sec. Do we want to let customers to pass Sync client too? If not, S3AsyncClient should be fine because S3CrtAsyncClient is also an instance of S3AsyncClient.

PR #194, opened 2024-12-11
A few minor optimisations 

ahmarsuhail:
## Description of change

This PR makes a few small changes:

1/ Allows for passing in content length if it is already known by the reader. This avoids an extra HEAD call.

2/ Merges consecutive ranges before passing it to physicalIO. When prefetching consecutive columns, we can end up with multiple small requests. In such cases, it is more efficient to merge these ranges into a single GET.

3/ Don't apply to sequential prefetch logic to LogicalIO plans that come from the parquet prefetcher. Here, since we already know the ranges we want for specific columns, we don't need to extend requests. When prefetching consecutive columns eg: [100-200, 201-300], sequential prefetching can cause some over reading. This is currently done by just checking if the Read is ASYNC, as the only place ASYNC reads happen currently are from the parquet prefetcher. This logic should be improved, see https://github.com/awslabs/analytics-accelerator-s3/issues/195 for more details.

#### Relevant issues

N/A

No

Yes, a new createStream() method that allows the passing of a contentLength. 

Ran unit tests, and local benchmarks.

- [ X] I have updated the CHANGELOG or README if appropriate

---

### DIFF
--- a/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/parquet/ParquetUtils.java
+++ b/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/parquet/ParquetUtils.java
@@ -143,4 +143,39 @@ public final class ParquetUtils {
 
     return rowGroupsToPrefetch;
   }
+
+  /**
+   * Merges consecutive ranges to avoid making multiple small requests. For example, if there are
+   * ranges [100-200, 500-600, 601-800, 801-900, 1000-1200], this list will be merged into [100-200,
+   * 500-900, 1000-1200]. This is a common scenario when reading smaller consecutive parquet
+   * columns, and when these consecutive columns are small (~1-2MB), making multiple GETs may,
+   * instead of a single larger merged request, may hurt performance.
+   *
+   * <p>This code can further be extended in the future, such that ranges that are not consecutive,
+   * but close enough together may also be merged. For example, ranges [100-200, 300-600] are not
+   * consecutive, but merging may be beneficial as the delta b/w them is only 100 bytes.
+   *
+   * @param ranges Range of requests to be merged
+   * @return merged ranges
+   */
+  public static List<Range> mergeRanges(List<Range> ranges) {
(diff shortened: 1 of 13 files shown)


input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/parquet/ParquetUtils.java
```diff
@@ -143,4 +143,39 @@ public static List<Integer> constructRowGroupsToPrefetch(ColumnMetadata columnMe
 
     return rowGroupsToPrefetch;
   }
+
+  /**
```
ahmarsuhail, line 147: Please review this carefully! While loops make me nervous

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/parquet/ParquetUtils.java
```diff
@@ -143,4 +143,39 @@ public static List<Integer> constructRowGroupsToPrefetch(ColumnMetadata columnMe
...
+   * instead of a single larger merged request, may hurt performance.
+   *
+   * <p>This code can further be extended in the future, such that ranges that are not consecutive,
+   * but close enough together may also be merged. For example, ranges [100-200, 300-600] are not
+   * consecutive, but merging may be beneficial as the delta b/w them is only 100 bytes.
+   *
+   * @param ranges Range of requests to be merged
+   * @return merged ranges
+   */
+  public static List<Range> mergeRanges(List<Range> ranges) {
```
CsengerG, line 161: I think this works as long as ranges are non-overlapping.  (I think it is, but...) ... is this guaranteed by the Parquet format?

CsengerG (approved):
LGTM. I had a small comment on how this works with nested Parquet.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/parquet/ParquetUtils.java
ahmarsuhail: should be fine! column ranges are never overlapping

Issue #201, opened 2024-12-18
Read with len = 0 should return 0 bytes read. 


fuatbasik:
### Analytics Accelerator Library for Amazon S3 Version

0.0.1, 0.0.2

### AWS Region

us-east-1

### Describe the running environment

Running in EC2 instance and mac. 

### What happened?

According to InputStream specification, when reader tries to read 0 bytes, read method should return 0 bytes read, independent of the object size. 
See:

```
If len is zero, then no bytes are read and 0 is returned; otherwise, there is an attempt to read at least one byte. If no byte is available because the stream is at end of file, the value -1 is returned; otherwise, at least one byte is read and stored into b.
```

In S3SeekableInputStream for an 0-sized object implementation of the read method will return -1.

https://github.com/awslabs/analytics-accelerator-s3/blob/main/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStream.java#L135-L153 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

fuatbasik:
It is fixed now.

PR #203, opened 2025-01-03
Fix Len = 0 and Insufficient Buffer behaviours for positioned reads

fuatbasik:
## Description of change

According to InputStream documentation Len = 0 should return 0

`If len is zero, then no bytes are read and 0 is returned;` 

Following the InputStream implementation, this commit is adding argument checks to Positioned Reads. This validation is implemented in the SeekableInputStream to ensure all subclasses behave the same way in the unhappy path.

Checks are done in the same order with InputStream implementation. 

#### Relevant issues

https://github.com/awslabs/analytics-accelerator-s3/issues/201

Yes. For a 0-length read to a 0-sized object read will now return 0 instead of -1.
Insufficient Buffer Capacity will now throw IndexOutOfBoundsException instead of IllegalArgumentException. 

No, it changes existing public API behaviour. 

Added new unit tests. 

- [N/A] I have updated the CHANGELOG or README if appropriate

---

### DIFF
--- a/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStream.java
+++ b/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStream.java
@@ -134,8 +134,11 @@ public class S3SeekableInputStream extends SeekableInputStream {
   @Override
   public int read(byte @NonNull [] buffer, int offset, int length) throws IOException {
     throwIfClosed("cannot read from closed stream");
+    validatePositionedReadArgs(position, buffer, offset, length);
 
-    if (this.position >= getContentLength()) {
+    if (length == 0) {
+      return 0;
+    } else if (this.position >= getContentLength()) {
       return EOF;
     }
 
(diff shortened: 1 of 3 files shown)


input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/SeekableInputStream.java
```diff
@@ -57,4 +58,38 @@ public abstract class SeekableInputStream extends InputStream {
...
+  protected void validatePositionedReadArgs(long position, byte[] buffer, int offset, int length) {
+    Preconditions.checkArgument(offset >= 0, "Offset is negative");
+    Preconditions.checkArgument(length >= 0, "Length is negative");
+
+    // TODO: S3A throws an EOFException here, S3FileIO does IllegalArgumentException
+    // TODO: https://github.com/awslabs/analytics-accelerator-s3/issues/84
+    Preconditions.checkArgument(position >= 0, "Position is negative");
+
+    if (buffer == null) {
+      throw new NullPointerException("Null destination buffer");
```
stubz151, line 83: why not just have both of these in precondition checks?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/SeekableInputStream.java
```diff
@@ -57,4 +58,38 @@ public abstract class SeekableInputStream extends InputStream {
...
+   *
+   * @param position the position to read from
+   * @param buffer the buffer to read into
+   * @param offset the offset in the buffer to start writing at
+   * @param length the number of bytes to read
+   * @throws IllegalArgumentException if the position, offset or length is negative
+   * @throws NullPointerException if the buffer is null
+   * @throws IndexOutOfBoundsException if the offset or length are invalid for the given buffer
+   */
+  protected void validatePositionedReadArgs(long position, byte[] buffer, int offset, int length) {
```
stubz151, line 74: maybe add a unit test for this method? should be quick

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStream.java
```diff
@@ -134,8 +134,11 @@ public int read() throws IOException {
   @Override
   public int read(byte @NonNull [] buffer, int offset, int length) throws IOException {
     throwIfClosed("cannot read from closed stream");
+    validatePositionedReadArgs(position, buffer, offset, length);
 
-    if (this.position >= getContentLength()) {
+    if (length == 0) {
+      return 0;
+    } else if (this.position >= getContentLength()) {
```
rajdchak, line 141: shall we move this validation as well to the method where we did the previous validations, so that all the validations are in same place?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/SeekableInputStream.java
fuatbasik: Makes perfect sense. doing it now.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStream.java
fuatbasik: This is a good recommendation. I did not move this validation because full content length might not be available to other streams those will implement this class. In S3 case we know the size thanks to object metadata. Let me know what you think.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/SeekableInputStream.java
fuatbasik: That makes perfect sense. Doing it now.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/SeekableInputStream.java
fuatbasik: Well i didn't because it is a protected method and not sure if we want to expose it. Instead i added tests to S3SeekableInputStream class which is publicly exposed. Let me know if you think this is a good approach.

stubz151 (approved):

PR #206, opened 2025-01-13
Add ability to dump configs

CsengerG:
## Description of change

This commit adds a `SerializableConfiguration` interface and makes use of it in the PhysicalIO, LogicalIO and Telemetry configurations.

Allowing clients to dump their stream config will allow them to verify whether their configurations get consumed as expected.

The approach taken here is a KISS -- adding Java reflection would complicate the code base and would involve a _major_ refactoring (I tried!). While I am convinced that we should at some point eliminate code duplication and simplify how we add new config parameters, it feels too expensive right now.

#### Relevant issues
N/A

No

No

- Runs on benchmark clusters

---

> By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and I agree to the terms of the [Developer Certificate of Origin (DCO)](https://developercertificate.org/).


stubz151 (approved):
[opinion] This might be better in json, just because there some nicer tools for working with json. And if this expands the formatting can come in handy.

common/src/main/java/software/amazon/s3/analyticsaccelerator/SerializableConfiguration.java
```diff
@@ -0,0 +1,27 @@
...
+/** An interface listing methods that a serialisable configuration should have. */
+public interface SerializableConfiguration {
+
+  /**
+   * Dump the key value pairs in this configuration object.
+   *
+   * @return a string representation of this configuration object.
+   */
+  String dumpConfig();
+}
```
rajdchak, line 27: Where and how do we want to use this interface?

rajdchak: I think I got the answer for this.

common/src/main/java/software/amazon/s3/analyticsaccelerator/SerializableConfiguration.java
```diff
@@ -0,0 +1,27 @@
...
+
+/** An interface listing methods that a serialisable configuration should have. */
+public interface SerializableConfiguration {
+
+  /**
+   * Dump the key value pairs in this configuration object.
+   *
+   * @return a string representation of this configuration object.
+   */
+  String dumpConfig();
```
rajdchak, line 26: Do we want to name this as dumpConfig or keep some name like clientConfig?

common/src/main/java/software/amazon/s3/analyticsaccelerator/common/telemetry/TelemetryConfiguration.java
```diff
@@ -110,4 +111,19 @@ public static TelemetryConfiguration fromConfiguration(
         .telemetryFormat(configuration.getString(TELEMETRY_FORMAT_KEY, DEFAULT_TELEMETRY_FORMAT))
         .build();
   }
+
+  @Override
+  public String dumpConfig() {
+    StringBuilder builder = new StringBuilder();
```
rajdchak, line 117: Nit: make this final here and in other places

common/src/main/java/software/amazon/s3/analyticsaccelerator/SerializableConfiguration.java
CsengerG: (Wanted to reply this to your previous comment which since got deleted.) Asking the "does this interface have a raison d'etre" is a really good point. I think the interface is probably superfluous, we should be overriding `@ToString` on the configuration classes rather than introducing yet another way to produce a human-readable view of this object.

Goes in hand with this comment of yours. Will address in the next rev.

CsengerG:
On Michael's comment: we talked offline and re-implemented JSON serialisation is probably out of scope for now and we definitely don't want to bring in external dependencies to handle it for something as small as this

rajdchak (approved):

PR #209, opened 2025-01-13
feat: adding etag checking for stream reads

stubz151:
## Description of change
We want etags on our get requests to verify that object reads are for the same version of the object as it is a durability concern. To do this I'm adding etags to the metadata for an object and then making conditional get requests.

Additionally pinning an objects metadata to a point in time. To stop us getting the metadata at different time points during a read. This idea helps us keep metadata consistent for a whole stream.

Additionally now am clearing the metadata for an object on exception. This will change when I pull in https://github.com/awslabs/analytics-accelerator-s3/pull/210. But just to demonstrate the idea I made it on all exceptions at first. 

Yes

Yes

* Unit tests:
For the unit tests: checking if we change the etag between blobs we get an error and used argument matcher for this
for this file [BlockTest.java](https://github.com/awslabs/analytics-accelerator-s3/compare/stubz151_etag?expand=1#diff-75cd3f62ba96f2b7d7e806e6b9c5572e63cbfb2e8eb1052ae02ca74ffa76cfb5) I just updated the values to include the etag don't think it's that important

* Integration tests:
For integration tests: added 2 one that verifies we fail when it changes mid read and a second that validates we pass and return the cached object if it's done reading.

TODO still, will update

---

### DIFF
--- a/common/src/main/java/software/amazon/s3/analyticsaccelerator/request/GetRequest.java
+++ b/common/src/main/java/software/amazon/s3/analyticsaccelerator/request/GetRequest.java
@@ -32,5 +32,5 @@ public class GetRequest {
   @NonNull S3URI s3Uri;
   @NonNull Range range;
   @NonNull Referrer referrer;
-  String etag;
+  @NonNull String etag;
 }
--- a/common/src/main/java/software/amazon/s3/analyticsaccelerator/request/GetRequest.java
+++ b/common/src/main/java/software/amazon/s3/analyticsaccelerator/request/GetRequest.java
@@ -30,4 +30,5 @@ public class GetRequest {
   @NonNull S3URI s3Uri;
   @NonNull Range range;
   @NonNull Referrer referrer;
+  String etag;
 }
--- a/common/src/main/java/software/amazon/s3/analyticsaccelerator/request/ObjectContent.java
+++ b/common/src/main/java/software/amazon/s3/analyticsaccelerator/request/ObjectContent.java
@@ -24,4 +24,5 @@ import lombok.Data;
 @Builder
 public class ObjectContent {
   InputStream stream;
+  String etag;
 }
--- a/common/src/main/java/software/amazon/s3/analyticsaccelerator/request/ObjectMetadata.java
+++ b/common/src/main/java/software/amazon/s3/analyticsaccelerator/request/ObjectMetadata.java
@@ -23,4 +23,5 @@ import lombok.Data;
 @Builder
 public class ObjectMetadata {
   long contentLength;
+  String etag;
 }
--- a/common/src/main/java/software/amazon/s3/analyticsaccelerator/request/ObjectMetadata.java
+++ b/common/src/main/java/software/amazon/s3/analyticsaccelerator/request/ObjectMetadata.java
@@ -24,5 +24,6 @@ import lombok.Data;
 @Builder
 public class ObjectMetadata {
   long contentLength;
-  Optional<String> etag;
+
+  @Builder.Default Optional<String> etag = Optional.empty();
 }
--- a/common/src/main/java/software/amazon/s3/analyticsaccelerator/request/ObjectMetadata.java
+++ b/common/src/main/java/software/amazon/s3/analyticsaccelerator/request/ObjectMetadata.java
@@ -15,6 +15,7 @@
  */
 package software.amazon.s3.analyticsaccelerator.request;
 
+import java.util.Optional;
 import lombok.Builder;
 import lombok.Data;
 
@@ -23,5 +24,5 @@ import lombok.Data;
 @Builder
 public class ObjectMetadata {
   long contentLength;
-  String etag;
+  Optional<String> etag;
 }
--- a/common/src/main/java/software/amazon/s3/analyticsaccelerator/request/ObjectMetadata.java
+++ b/common/src/main/java/software/amazon/s3/analyticsaccelerator/request/ObjectMetadata.java
@@ -15,9 +15,9 @@
  */
 package software.amazon.s3.analyticsaccelerator.request;
 
-import java.util.Optional;
 import lombok.Builder;
 import lombok.Data;
+import lombok.NonNull;
 
 /** Wrapper class around HeadObjectResponse abstracting away from S3-specific details */
 @Data
@@ -25,5 +25,5 @@ import lombok.Data;
 public class ObjectMetadata {
   long contentLength;
 
-  @Builder.Default Optional<String> etag = Optional.empty();
+  @NonNull String etag;
 }
--- a/common/src/main/java/software/amazon/s3/analyticsaccelerator/util/ObjectKey.java
+++ b/common/src/main/java/software/amazon/s3/analyticsaccelerator/util/ObjectKey.java
@@ -15,7 +15,7 @@
  */
 package software.amazon.s3.analyticsaccelerator.util;
 
-import java.util.Objects;
+import java.util.Optional;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Getter;
@@ -27,7 +27,8 @@ import lombok.NonNull;
 @Builder
 public class ObjectKey {
   @NonNull S3URI s3URI;
-  @NonNull String etag;
+
+  Optional<String> etag;
 
   @Override
   public boolean equals(Object o) {
@@ -40,6 +41,13 @@ public class ObjectKey {
 
   @Override
   public int hashCode() {
-    return Objects.hash(s3URI, etag);
+    int result = s3URI.hashCode();
+
+    // Include etag in hash only if present
+    if (etag.isPresent()) {
+      result = result + etag.hashCode();
+    }
+
+    return result;
   }
 }
--- a/common/src/main/java/software/amazon/s3/analyticsaccelerator/util/ObjectKey.java
+++ b/common/src/main/java/software/amazon/s3/analyticsaccelerator/util/ObjectKey.java
@@ -13,24 +13,33 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package software.amazon.s3.analyticsaccelerator.request;
+package software.amazon.s3.analyticsaccelerator.util;
 
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import java.util.Objects;
+import lombok.AllArgsConstructor;
 import lombok.Builder;
+import lombok.Getter;
 import lombok.NonNull;
-import lombok.Value;
-import software.amazon.s3.analyticsaccelerator.util.S3URI;
 
-/**
- * Object representing arguments to a GetObject call. This class helps us abstract away from S3 SDK
- * constructs.
- */
-@Value
+/** Container used to represent an S3 object for a specific version/etag */
+@Getter
+@AllArgsConstructor
 @Builder
-@SuppressFBWarnings(value = "NM_CONFUSING", justification = "Sharing Getter names is not confusing")
-public class GetRequest {
-  @NonNull S3URI s3Uri;
-  @NonNull Range range;
-  @NonNull Referrer referrer;
-  String etag;
+public class ObjectKey {
+  @NonNull S3URI s3URI;
+  @NonNull String etag;
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
+
+    ObjectKey objectKey = (ObjectKey) o;
+    return s3URI.equals(objectKey.s3URI) && etag.equals(objectKey.etag);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(s3URI, etag);
+  }
 }
--- a/common/src/main/java/software/amazon/s3/analyticsaccelerator/util/ObjectKey.java
+++ b/common/src/main/java/software/amazon/s3/analyticsaccelerator/util/ObjectKey.java
@@ -15,7 +15,6 @@
  */
 package software.amazon.s3.analyticsaccelerator.util;
 
-import java.util.Optional;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Getter;
@@ -27,8 +26,7 @@ import lombok.NonNull;
 @Builder
 public class ObjectKey {
   @NonNull S3URI s3URI;
-
-  @Builder.Default Optional<String> etag = Optional.empty();
+  @NonNull String etag;
 
   @Override
   public boolean equals(Object o) {
@@ -41,13 +39,6 @@ public class ObjectKey {
 
   @Override
   public int hashCode() {
-    int result = s3URI.hashCode();
-
-    // Include etag in hash only if present
-    if (etag.isPresent()) {
-      result = result + etag.hashCode();
-    }
-
-    return result;
+    return s3URI.hashCode() + etag.hashCode();
   }
 }
--- a/common/src/test/java/software/amazon/s3/analyticsaccelerator/util/ObjectKeyTest.java
+++ b/common/src/test/java/software/amazon/s3/analyticsaccelerator/util/ObjectKeyTest.java
@@ -20,7 +20,6 @@ import static org.junit.jupiter.api.Assertions.*;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Optional;
 import org.junit.jupiter.api.Test;
 
 @SuppressFBWarnings(
@@ -35,10 +34,10 @@ public class ObjectKeyTest {
     S3URI s3uri2 = S3URI.of("bucket", "key");
     S3URI s3uri3 = S3URI.of("bucket", "key2");
 
-    ObjectKey key1 = ObjectKey.builder().s3URI(s3uri1).etag(Optional.of("etag1")).build();
-    ObjectKey key2 = ObjectKey.builder().s3URI(s3uri2).etag(Optional.of("etag1")).build();
-    ObjectKey key3 = ObjectKey.builder().s3URI(s3uri1).etag(Optional.of("etag2")).build();
-    ObjectKey key4 = ObjectKey.builder().s3URI(s3uri3).etag(Optional.of("etag1")).build();
+    ObjectKey key1 = ObjectKey.builder().s3URI(s3uri1).etag("etag1").build();
+    ObjectKey key2 = ObjectKey.builder().s3URI(s3uri2).etag("etag1").build();
+    ObjectKey key3 = ObjectKey.builder().s3URI(s3uri1).etag("etag2").build();
+    ObjectKey key4 = ObjectKey.builder().s3URI(s3uri3).etag("etag1").build();
 
     Map<ObjectKey, String> map = new HashMap<>();
 
@@ -68,10 +67,10 @@ public class ObjectKeyTest {
     S3URI s3uri2 = S3URI.of("bucket", "key");
     S3URI s3uri3 = S3URI.of("bucket", "key2");
 
-    ObjectKey key1 = ObjectKey.builder().s3URI(s3uri1).etag(Optional.of("etag1")).build();
-    ObjectKey key2 = ObjectKey.builder().s3URI(s3uri2).etag(Optional.of("etag1")).build();
-    ObjectKey key3 = ObjectKey.builder().s3URI(s3uri1).etag(Optional.of("etag2")).build();
-    ObjectKey key4 = ObjectKey.builder().s3URI(s3uri3).etag(Optional.of("etag1")).build();
+    ObjectKey key1 = ObjectKey.builder().s3URI(s3uri1).etag("etag1").build();
+    ObjectKey key2 = ObjectKey.builder().s3URI(s3uri2).etag("etag1").build();
+    ObjectKey key3 = ObjectKey.builder().s3URI(s3uri1).etag("etag2").build();
+    ObjectKey key4 = ObjectKey.builder().s3URI(s3uri3).etag("etag1").build();
 
     // Test equals
     assertEquals(key1, key1); // Same object
@@ -95,20 +94,21 @@ public class ObjectKeyTest {
     assertThrows(
         NullPointerException.class,
         () -> {
-          ObjectKey.builder().s3URI(null).etag(Optional.of("etag1")).build();
+          ObjectKey.builder().s3URI(null).etag("etag1").build();
         });
 
     // Test null etag
-    assertDoesNotThrow(
+    assertThrows(
+        NullPointerException.class,
         () -> {
-          ObjectKey.builder().s3URI(s3uri).etag(Optional.empty()).build();
+          ObjectKey.builder().s3URI(s3uri).etag(null).build();
         });
 
     // Test constructor with null values
     assertThrows(
         NullPointerException.class,
         () -> {
-          new ObjectKey(null, Optional.of("etag1"));
+          new ObjectKey(null, null);
         });
   }
 }
--- a/common/src/test/java/software/amazon/s3/analyticsaccelerator/util/ObjectKeyTest.java
+++ b/common/src/test/java/software/amazon/s3/analyticsaccelerator/util/ObjectKeyTest.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package software.amazon.s3.analyticsaccelerator.util;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+@SuppressFBWarnings(
+    value = "NP_NONNULL_PARAM_VIOLATION",
+    justification = "We mean to pass nulls to checks")
+public class ObjectKeyTest {
+
+  @Test
+  void testMapOperations() {
+    // Create test objects
+    S3URI s3uri1 = S3URI.of("bucket", "key");
+    S3URI s3uri2 = S3URI.of("bucket", "key");
+    S3URI s3uri3 = S3URI.of("bucket", "key2");
+
+    ObjectKey key1 = ObjectKey.builder().s3URI(s3uri1).etag("etag1").build();
+
+    ObjectKey key2 = ObjectKey.builder().s3URI(s3uri2).etag("etag1").build();
+
+    ObjectKey key3 = ObjectKey.builder().s3URI(s3uri1).etag("etag2").build();
+
+    ObjectKey key4 = ObjectKey.builder().s3URI(s3uri3).etag("etag1").build();
+
+    Map<ObjectKey, String> map = new HashMap<>();
+
+    // Test putting and retrieving
+    map.put(key1, "value1");
+    assertEquals("value1", map.get(key1));
+
+    // Test same s3URI and etag should be treated as same key
+    map.put(key2, "value2");
+    assertEquals("value2", map.get(key1));
+    assertEquals(1, map.size());
+
+    // Test different etag should be treated as different key
+    map.put(key3, "value3");
+    assertEquals("value3", map.get(key3));
+    assertEquals(2, map.size());
+
+    // Test different s3URI should be treated as different key
+    map.put(key4, "value4");
+    assertEquals("value4", map.get(key4));
+    assertEquals(3, map.size());
+  }
+
+  @Test
+  void testEqualsAndHashCodeObjectKey() {
+    S3URI s3uri1 = S3URI.of("bucket", "key");
+    S3URI s3uri2 = S3URI.of("bucket", "key");
+    S3URI s3uri3 = S3URI.of("bucket", "key2");
+
+    ObjectKey key1 = ObjectKey.builder().s3URI(s3uri1).etag("etag1").build();
+
+    ObjectKey key2 = ObjectKey.builder().s3URI(s3uri2).etag("etag1").build();
+
+    ObjectKey key3 = ObjectKey.builder().s3URI(s3uri1).etag("etag2").build();
+
+    ObjectKey key4 = ObjectKey.builder().s3URI(s3uri3).etag("etag1").build();
+
+    // Test equals
+    assertEquals(key1, key1); // Same object
+    assertEquals(key1, key2); // Same values
+    assertNotEquals(key1, key3); // Different etag
+    assertNotEquals(key1, key4); // Different s3URI
+    assertNotEquals(key1, null); // Null comparison
+    assertNotEquals(key1, "not an ObjectKey"); // Different type
+
+    // Test hashCode
+    assertEquals(key1.hashCode(), key2.hashCode()); // Same values should have same hash
+    assertNotEquals(key1.hashCode(), key3.hashCode()); // Different etag should have different hash
+    assertNotEquals(key1.hashCode(), key4.hashCode()); // Different s3URI should have different hash
+  }
+
+  @Test
+  void testNullValidation() {
+    S3URI s3uri = S3URI.of("bucket", "key");
+
+    // Test null s3URI
+    assertThrows(
+        NullPointerException.class,
+        () -> {
+          ObjectKey.builder().s3URI(null).etag("etag1").build();
+        });
+
+    // Test null etag
+    assertThrows(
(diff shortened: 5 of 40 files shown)


common/src/main/java/software/amazon/s3/analyticsaccelerator/request/GetRequest.java
```diff
@@ -30,4 +30,5 @@ public class GetRequest {
   @NonNull S3URI s3Uri;
   @NonNull Range range;
   @NonNull Referrer referrer;
+  String etag;
```
CsengerG, line 33: Can it ever be null?

doc/DEVELOPMENT.md
```diff
@@ -23,6 +23,7 @@ To generate data and run benchmarks, you first need to configure your environmen
 
 ### Data Generation
 After your environment is configured, you can generate data to run benchmarks against.
+* If you haven't already done so create a new S3 Bucket and add the bucket name(`S3_TEST_BUCKET`) and prefix(`S3_TEST_PREFIX`) as env vars
```
CsengerG, line 26: Thanks!

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/ConcurrencyCorrectnessTest.java
```diff
@@ -63,6 +63,26 @@ void testQuasiParquetReads(
         s3Object, streamReadPattern, configuration, CONCURRENCY_LEVEL, CONCURRENCY_ITERATIONS);
   }
 
+  @ParameterizedTest
+  @MethodSource("etagTests")
+  void testChangingEtagFailsStream(
+      S3Object s3Object,
+      StreamReadPatternKind streamReadPattern,
+      DATInputStreamConfigurationKind configuration)
```
CsengerG, line 71: If we are already here, can we get rid of references of "DAT"? (This is the old name we should have scrapped before; sorry to scope creep.)

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/IntegrationTestBase.java
```diff
@@ -92,6 +103,143 @@ protected void testAndCompareStreamReadPattern(
...
+
+      // Change the file
+      s3Client
+          .putObject(
+              x -> x.bucket(s3URI.getBucket()).key(s3URI.getKey()),
+              AsyncRequestBody.fromBytes(generateRandomBytes(bufferSize)))
+          .join();
+
+      // read the next bytes and fail.
+      Exception ex =
```
CsengerG, line 149: Can we say more about the type of exception it throws? Do we expect IOException or something else?

stubz151:
One thing I'm not sure about is recovery here. If the stream fails half way should I be clearing this data from the BlobStore/Metadata Store? or no?

object-client/src/test/java/software/amazon/s3/analyticsaccelerator/S3SdkObjectClientTest.java
```diff
@@ -162,8 +172,37 @@ void testGetObjectWithRange() {
...
+                  .referrer(new Referrer("bytes=0-20", ReadMode.SYNC))
+                  .build()));
+    }
+  }
+
+  @Test
+  void testGetObjectWithDifferentEtagsThrowsError() {
+    try (S3AsyncClient s3AsyncClient = createMockClient()) {
+      S3SdkObjectClient client = new S3SdkObjectClient(s3AsyncClient);
+      assertInstanceOf(
```
CsengerG, line 185: What do we gain from this assert? Is this asserting that the result is not null or why do we have it?

object-client/src/test/java/software/amazon/s3/analyticsaccelerator/S3SdkObjectClientTest.java
```diff
@@ -249,15 +288,46 @@ private static S3AsyncClient createMockClient() {
...
+            argThat(
+                (ArgumentMatcher<GetObjectRequest>)
+                    request -> {
+                      if (request == null) {
+                        return false;
+                      }
+                      return (request).ifMatch() != null && !(request).ifMatch().equals(ETAG);
+                    }),
+            (AsyncResponseTransformer<GetObjectResponse, Object>) any()))
+        .thenThrow(S3Exception.builder().message("PreconditionFailed").statusCode(412).build());
```
CsengerG, line 329: There is some duplication here. This is test code so maybe not crucial, but is there a way to simplify this, in case we need to extend further in the future?

CsengerG (commented):
Thanks a lot for this! This is very important to get right.

Had minor questions around code style.

CsengerG:
> One thing I'm not sure about is recovery here. If the stream fails half way should I be clearing this data from the BlobStore/Metadata Store? or no?

Not sure about the metadata store, but as far as durability goes, there is no point storing incoherent parts of an object in the cache.  Is it easy to evict the whole Blob once we detect it has changed?

common/src/main/java/software/amazon/s3/analyticsaccelerator/request/GetRequest.java
stubz151: Was thinking for your first getRequest this could be null. In our case we do a headrequest first so we have this but this isn't necessarily always going to be true.

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/IntegrationTestBase.java
stubz151: It's a normal exception with the cause being an s3 exception it gets wrapped so we sort of have to unpack it.

object-client/src/test/java/software/amazon/s3/analyticsaccelerator/S3SdkObjectClientTest.java
stubz151: I thought it was a way of asserting that we not throwing an error for all etags only when it's different to the one we expect, but you can tell me what you think.

object-client/src/test/java/software/amazon/s3/analyticsaccelerator/S3SdkObjectClientTest.java
stubz151: thought about this a lot actually, moving everything to a base test class for all unit tests (like we have for our integration tests) makes a lot of sense here. Didn't want to scope creep myself too much and make a massive pr, but can add a backlog item to just share some common constants/functions between tests.

stubz151:
> > One thing I'm not sure about is recovery here. If the stream fails half way should I be clearing this data from the BlobStore/Metadata Store? or no?
> 
> Not sure about the metadata store, but as far as durability goes, there is no point storing incoherent parts of an object in the cache. Is it easy to evict the whole Blob once we detect it has changed?

I don't think it is easy from what I can see but if it's something we need to be doing I will go and check what options we have.

fuatbasik (commented):
Thanks for the PR @stubz151 . I think this is covering the lowest level of getting stale data from S3 but it does not protect BlockManager to serve stale data. Consider the case entire Object is in memory already but we know the etag has changed. A new stream open will still serve the bytes of the old object.

To solve this, we probably need to start keeping the etag at the blob level and pass it to every read request, whenever we are passing an s3URI today. Also put some other comments

fuatbasik (changes requested):
Please see me previous comments.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManager.java
```diff
@@ -177,6 +177,9 @@ public synchronized void makeRangeAvailable(long pos, long len, ReadMode readMod
 
     // Fix "effectiveEnd", so we can pass it into the lambda
     final long effectiveEndFinal = effectiveEnd;
+
+    String etag = this.metadataStore.get(s3URI).getEtag();
```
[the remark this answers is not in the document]
stubz151: so we do a compute if absent so if we fail while computing that it will fail and we will throw that which I think is what we want here.

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/IntegrationTestBase.java
```diff
@@ -92,6 +103,143 @@ protected void testAndCompareStreamReadPattern(
...
+          assertThrows(Exception.class, () -> readAndAssert(stream, buffer, 200, readAheadBytes));
+      S3Exception s3Exception =
```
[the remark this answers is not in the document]
stubz151: This is dealt with in https://github.com/awslabs/analytics-accelerator-s3/pull/210

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/IntegrationTestBase.java
```diff
@@ -92,6 +103,143 @@ protected void testAndCompareStreamReadPattern(
...
+   * @param DATInputStreamConfigurationKind configuration kind
+   */
+  protected void testChangingEtagMidStream(
+      @NonNull S3Object s3Object,
+      @NonNull DATInputStreamConfigurationKind DATInputStreamConfigurationKind)
+      throws IOException {
+    int bufferSize = (int) s3Object.getSize();
+    byte[] buffer = new byte[bufferSize];
+
+    // Create the s3DATClientStreamReader - that creates the shared state
```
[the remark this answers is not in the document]
stubz151: Cutting an issue to refactor our unit tests to remove mentions of dat and to move some of the shared code to helper classes we happy with that? thought this change was already too big.

common/src/test/java/software/amazon/s3/analyticsaccelerator/util/ObjectKeyTest.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+@SuppressFBWarnings(
+    value = "NP_NONNULL_PARAM_VIOLATION",
+    justification = "We mean to pass nulls to checks")
+public class ObjectKeyTest {
+
+  @Test
+  void testMapOperations() {
```
stubz151, line 31: added this because I got caught by it initially and didn't realize my hash wasn't functioning as expected till later on. So wanted  to make sure we had validation for that.

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/ConcurrencyCorrectnessTest.java
stubz151: Cutting an issue for this.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/impl/PhysicalIOImpl.java
```diff
@@ -76,11 +78,13 @@ public PhysicalIOImpl(
       @NonNull BlobStore blobStore,
       @NonNull Telemetry telemetry,
       StreamContext streamContext) {
-    this.s3URI = s3URI;
     this.metadataStore = metadataStore;
     this.blobStore = blobStore;
     this.telemetry = telemetry;
     this.streamContext = streamContext;
+
+    this.metadata = this.metadataStore.get(s3URI);
```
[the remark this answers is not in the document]
CsengerG: It is also OK if we cannot evaluate impact in theory -- a benchmark run might help, but we are changing how we time HEAD requests as far as I can tell.

common/src/main/java/software/amazon/s3/analyticsaccelerator/request/GetRequest.java
CsengerG: Might consider using an Optional, but I am not too opinionated.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStreamFactory.java
```diff
@@ -96,13 +92,14 @@ public S3SeekableInputStream createStream(@NonNull S3URI s3URI) {
    *
    * @param s3URI the object's S3 URI
    * @param contentLength content length
+   * @param etag etag for content
    * @return An instance of the input stream.
    */
-  public S3SeekableInputStream createStream(@NonNull S3URI s3URI, long contentLength) {
+  public S3SeekableInputStream createStream(@NonNull S3URI s3URI, long contentLength, String etag) {
```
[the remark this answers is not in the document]
fuatbasik: +1 on this. This doesn't seem really intuitive to me. We should double check precedence here. If we skip getting etag here, can we actually fill it with the first Get request?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Blob.java
```diff
@@ -25,36 +25,37 @@
...
-      @NonNull S3URI s3URI,
-      @NonNull MetadataStore metadataStore,
+      @NonNull ObjectKey objectKey,
+      @NonNull ObjectMetadata metadata,
       @NonNull BlockManager blockManager,
       @NonNull Telemetry telemetry) {
 
-    this.s3URI = s3URI;
-    this.metadataStore = metadataStore;
+    this.s3URI = objectKey;
```
fuatbasik, line 57: shall we rename s3URI to ObjectKey?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Blob.java
```diff
@@ -149,7 +151,7 @@ public IOPlanExecution execute(IOPlan plan) {
   }
 
   private long contentLength() {
-    return metadataStore.get(s3URI).getContentLength();
+    return metadata.getContentLength();
```
fuatbasik, line 154: IIRC, getting the contentLength was async and evaluated lazily only when we actually need it. With this change are we making it eager?

fuatbasik (commented):
@stubz151  Put some comments but i need to have another look when i have more time, especially around how this change is refactoring MetadataStore lookups.

common/src/test/java/software/amazon/s3/analyticsaccelerator/util/ObjectKeyTest.java
stubz151: So I originally didn't override the hash function and thought the default would be enough, but I picked it up while running the integ tests that the blobstore was just growing with every blobstore.get

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStreamFactory.java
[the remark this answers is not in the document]
stubz151: We can easily fill this yes, happy to just change it to do that.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/impl/PhysicalIOImpl.java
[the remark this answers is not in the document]
stubz151: I think I was worried about passing a commutable future as an argument and if it could lead to a situation where metadata is different but if we don't think that could happen I'm happy to test, think it was just to help my mental model while coding. 

I'm not sure about the impact here, I think for every stream on a new uri we now immediately doing a HEAD request versus only lazily doing it later on, in theory it's still just one head request just at different times.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/impl/PhysicalIOImpl.java
```diff
@@ -183,17 +207,29 @@ public int readTail(byte[] buf, int off, int len) throws IOException {
...
+    } catch (Exception e) {
+      handleOperationExceptions(e);
+      throw e;
+    }
+  }
+
+  private void handleOperationExceptions(Exception e) {
+    System.out.println("Exception here: evicting bad key" + e);
+    metadataStore.evictKey(this.objectKey.getS3URI());
+    blobStore.evictKey(this.objectKey);
```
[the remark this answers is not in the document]
stubz151: I wanted to extend this a little but ideally a cleanup shouldn't fail a stream and shouldn't fail if one fails. If we want I can just put both in futures and execute them so we don't care about the result and they independent.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Blob.java
stubz151: ugh missed this one but yes, was meant to rename all of them.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Blob.java
stubz151: This would be eager, as mentioned on the previous thread for this while coding it was sort of to just help my mental model but am happy to change it back to a future if we worried about performance.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/impl/PhysicalIOImpl.java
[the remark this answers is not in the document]
rajdchak: I think for parquet data formats we also have a parquetColumnPrefetchStore, do we also want to clean that?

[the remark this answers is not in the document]
stubz151: I don't think so because we are pre-fetching that data, but I will try run some tests and see what I can find. Can you imagine a scenario that could go wrong here?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/impl/PhysicalIOImpl.java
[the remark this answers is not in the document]
CsengerG: > it's still just one head request just at different times

This is true, as long as the engine doesn't open a file but then does not read anything (because then the issued HEADs are useless work) -- again, I don't know if this happens or not.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStreamFactory.java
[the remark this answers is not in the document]
stubz151: I thought it was a bit weird where if we not specifying the etag and we do a headobject and the content length has changed what do we do in that scenario. If we respect the original content length then we will build an incomplete blob. If we get a new one then passing it through seems pointless?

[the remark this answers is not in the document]
ahmarsuhail: I'm just catching up on things and trying to get my brain to remember what this library does! I added this in for EMR, where you have the HEAD response already, and that is used to build S3AFileStatus, which gets passed in when creating a new stream, and so avoids the new head. 


Here is the the class in S3A: https://github.com/apache/hadoop/blob/trunk/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileStatus.java

At that time, the only requirement was for contentLen. What we should do imo, is create a similar container class in our library, that can store etag and content len, and update the constructor to take that value. 

On Geza's second point, S3A has `fs.s3a.change.detection.mode`, which basically controls this. See https://github.com/apache/hadoop/blob/trunk/hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/index.md for more details.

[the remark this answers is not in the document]
stubz151: We have a container for objectMetadata that has content len, should we just add a constructor for that?

And then I can add in a config like this.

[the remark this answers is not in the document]
ahmarsuhail: works for me, will just need to update the EMR integration code once this is in merged in to use that as well.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/impl/PhysicalIOImpl.java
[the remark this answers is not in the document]
ahmarsuhail: This was never really lazy to begin with, as all the prefetching logic would kick in as soon as you opened a new stream, and that logic needs the content len straight away.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/impl/PhysicalIOImpl.java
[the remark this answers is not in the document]
ahmarsuhail: It would be good to evict from there, but not necessary for correctness. We store only data that helps us figure out what ranges to prefetch there, so if the file changes, we'll get prefetched ranges wrong. But when the actual reads come in, we'll get the right data synchronously. 

We could do that as a follow up?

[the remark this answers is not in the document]
stubz151: For the exception stuff, I was waiting for Sanjay's pr to add some exceptions to be merged and yeah will move it to the specific exceptions in this pr.

I think lets add a backlog item and do it as a follow up, there is already a lot of changes here.

common/src/main/java/software/amazon/s3/analyticsaccelerator/request/ObjectContent.java
```diff
@@ -24,4 +24,5 @@
 @Builder
 public class ObjectContent {
   InputStream stream;
+  String etag;
```
ahmarsuhail, line 27: why do you need etag here?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
```diff
@@ -33,6 +33,7 @@ public class PhysicalIOConfiguration {
   private static final int DEFAULT_CAPACITY_BLOB_STORE = 50;
   private static final int DEFAULT_CAPACITY_METADATA_STORE = 50;
   private static final boolean DEFAULT_USE_SINGLE_CACHE = true;
+  private static final boolean DEFAULT_DETECTION_MODE_ON = true;
```
ahmarsuhail, line 36: nit: rename to DEFAULT_CHANGE_DETECTION_MODE

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
```diff
@@ -93,6 +94,14 @@ public class PhysicalIOConfiguration {
 
   private static final String SEQUENTIAL_PREFETCH_SPEED_KEY = "sequentialprefetch.speed";
 
+  /**
+   * Whether to check etags for object. {@link PhysicalIOConfiguration#DEFAULT_DETECTION_MODE_ON} by
+   * default.
+   */
+  @Builder.Default private boolean detectionModeOn = DEFAULT_DETECTION_MODE_ON;
+
+  private static final String DETECTION_MODE_ON = "detectionmodeon";
```
ahmarsuhail, line 103: nit: key should be `change.detection` or similar

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
```diff
@@ -110,6 +110,17 @@ public Block(
     this.objectKey = objectKey;
     this.range = new Range(start, end);
 
+    GetRequest.GetRequestBuilder getRequestBuilder =
+        GetRequest.builder()
+            .s3Uri(this.objectKey.getS3URI())
+            .range(this.range)
+            .referrer(new Referrer(range.toHttpString(), readMode));
+
+    if (this.objectKey.getEtag().isPresent()) {
```
ahmarsuhail, line 119: shouldn't you only add the etag to the getRequest if change detection is on?

ahmarsuhail (changes requested):

common/src/main/java/software/amazon/s3/analyticsaccelerator/request/ObjectContent.java
stubz151: don't was testing stuff and this slipped in :(

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
stubz151: Well this should not be present if change detection is off can also add the check here if we want to have some guard rails

stubz151: spoke about this in person and we happy with this approach being the most elegant for now.

input-stream/build.gradle.kts
```diff
@@ -67,6 +67,7 @@ dependencies {
     api(project(":object-client"))
 
     implementation(project(":common"))
+    implementation(libs.s3);
```
stubz151, line 70: adding this in so we can listen to s3 exceptions, happy to remove it and look for the error code or message but I felt like we needed to add this eventually our jmh and tests already rely on it here.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStreamFactory.java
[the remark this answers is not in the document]
stubz151: One option is to keep the old content len constructor as well just to help the swap between the 2 what do we think?

input-stream/build.gradle.kts
fuatbasik: input-stream should not depend on the underlying storage layer. For exceptions we should use `the error code or message` as you described. For JMH and tests, we should consider how to move them out of input-stream if they are s3-specific.

common/src/main/java/software/amazon/s3/analyticsaccelerator/util/ObjectKey.java
```diff
@@ -0,0 +1,53 @@
...
+import lombok.NonNull;
+
+/** Container used to represent an S3 object for a specific version/etag */
+@Getter
+@AllArgsConstructor
+@Builder
+public class ObjectKey {
+  @NonNull S3URI s3URI;
+
+  @Builder.Default Optional<String> etag = Optional.empty();
```
fuatbasik, line 31: nit: Can we avoid keeping etag in ObjectKey and in ObjectMetadata?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStreamFactory.java
```diff
@@ -87,22 +83,22 @@ public S3SeekableInputStreamFactory(
    * @param s3URI the object's S3 URI
    * @return An instance of the input stream.
    */
-  public S3SeekableInputStream createStream(@NonNull S3URI s3URI) {
+  public S3SeekableInputStream createStream(@NonNull S3URI s3URI) throws IOException {
```
ahmarsuhail, line 86: why does this throw an IoException now?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
```diff
@@ -93,6 +94,14 @@ public class PhysicalIOConfiguration {
 
   private static final String SEQUENTIAL_PREFETCH_SPEED_KEY = "sequentialprefetch.speed";
 
+  /**
+   * Whether to check etags for object. {@link
+   * PhysicalIOConfiguration#DEFAULT_CHANGE_DETECTION_MODE} by default.
+   */
+  @Builder.Default private boolean detectionMode = DEFAULT_CHANGE_DETECTION_MODE;
```
ahmarsuhail, line 101: sorry for asking for a name change again, but if we're only going with a boolean, then can we have DEFAULT_CHANGE_DETECTION__ENABLED, instead of mode?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
```diff
@@ -93,6 +94,14 @@ public class PhysicalIOConfiguration {
...
+
+  private static final String DETECTION_MODE_KEY = "change.detection";
```
ahmarsuhail, line 103: same here, change.detection.enabled

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
```diff
@@ -33,6 +33,7 @@ public class PhysicalIOConfiguration {
   private static final int DEFAULT_CAPACITY_BLOB_STORE = 50;
   private static final int DEFAULT_CAPACITY_METADATA_STORE = 50;
   private static final boolean DEFAULT_USE_SINGLE_CACHE = true;
+  private static final boolean DEFAULT_CHANGE_DETECTION_MODE = true;
```
fuatbasik, line 36: i do not think we should make this configurable.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
stubz151: sure.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
ahmarsuhail: sorry this is my bad again, if you're in Block.java we can get access to physicalConfig, we should be able to pass it down, and then use the config value.

common/src/main/java/software/amazon/s3/analyticsaccelerator/request/ObjectMetadata.java
```diff
@@ -23,4 +24,6 @@
 @Builder
 public class ObjectMetadata {
   long contentLength;
+
+  @Builder.Default Optional<String> etag = Optional.empty();
```
fuatbasik, line 28: we discussed this offline, lets make ETag required.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockStore.java
```diff
@@ -24,29 +24,30 @@
...
-   * @param metadataStore the metadata cache
+   * @param objectKey the etag and S3 URI of the object
+   * @param metadata the metadata for the object
    */
-  public BlockStore(S3URI s3URI, MetadataStore metadataStore) {
-    Preconditions.checkNotNull(s3URI, "`s3URI` must not be null");
-    Preconditions.checkNotNull(metadataStore, "`metadataStore` must not be null");
+  public BlockStore(ObjectKey objectKey, ObjectMetadata metadata) {
+    Preconditions.checkNotNull(objectKey, "`objectKey` must not be null");
+    Preconditions.checkNotNull(metadata, "`metadataStore` must not be null");
```
fuatbasik, line 47: nit: error message still mentions metadataStore

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStreamFactory.java
stubz151: chatted about this in person but it's because we getting the metadata at the start now.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
stubz151: ended up deleting this

common/src/main/java/software/amazon/s3/analyticsaccelerator/util/ObjectKey.java
stubz151: Not really they 2 separate entities in our caches and we need the etag in metadata to get the latest value. And we need it in the blob to make sure we return the correct objects for the cache.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
stubz151: [done]

common/src/main/java/software/amazon/s3/analyticsaccelerator/request/ObjectMetadata.java
stubz151: [done]

fuatbasik (approved):
Thanks a lot @stubz151 . I am happy with these changes. I really appreciate all the effort!

PR #210, opened 2025-01-15
Improve the exception handling of the S3SdkObjectClient

SanjayMarreddi:
## Description of change
This PR Improves the exception handling of the S3SdkObjectClient. 

#### TODO before merging
- [x] Improve the documentation of the methods modified.
- [x] Add few more unit and integration tests in this library as appropriate.
- [x] Test with the `hadoop-aws` integration tests and ensure previously failing Exception related tests pass now.

No

No

- Integration and unit tests.

---

### DIFF
--- a/common/src/main/java/software/amazon/s3/analyticsaccelerator/common/telemetry/DefaultTelemetry.java
+++ b/common/src/main/java/software/amazon/s3/analyticsaccelerator/common/telemetry/DefaultTelemetry.java
@@ -15,8 +15,11 @@
  */
 package software.amazon.s3.analyticsaccelerator.common.telemetry;
 
+import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 import lombok.*;
@@ -261,6 +264,55 @@ public class DefaultTelemetry implements Telemetry {
                 .value(value)
                 .kind(MetricMeasurementKind.RAW)
                 .build());
+  }
+
+  /**
+   * This is a helper method to reduce verbosity on completed futures. Blocks on the execution on
+   * {@link CompletableFuture#join()} and records the telemetry as {@link Operation}. We do not
+   * currently carry the operation into the context of any continuations, so any {@link Operation}s
+   * that are created in that context need to carry the parenting chain. The telemetry is only
+   * recorded if the future was not completed, which is checked via {@link
+   * CompletableFuture#isDone()}
+   *
+   * @param <T> - return type of the {@link CompletableFuture<T>}.
+   * @param level telemetry level.
+   * @param operationSupplier operation to record this execution as.
+   * @param operationCode the future to measure the execution of.
+   * @return an instance of {@link T} that returns the same result as the one passed in.
+   * @throws IOException if the underlying operation threw an IOException
+   */
+  @Override
+  public <T> T measureJoin(
+      @NonNull TelemetryLevel level,
+      @NonNull OperationSupplier operationSupplier,
+      @NonNull CompletableFuture<T> operationCode)
+      throws IOException {
+    if (operationCode.isDone()) {
+      return handleCompletableFutureJoin(operationCode);
+    } else {
+      return this.measure(
+          level, operationSupplier, () -> handleCompletableFutureJoin(operationCode));
+    }
+  }
+
+  /**
+   * Helper method to handle CompletableFuture join() operation and properly unwrap exceptions
+   *
+   * @param <T> - return type of the CompletableFuture
+   * @param future the CompletableFuture to join
+   * @return the result of the CompletableFuture
+   * @throws IOException if the underlying future threw an IOException
+   */
+  private <T> T handleCompletableFutureJoin(CompletableFuture<T> future) throws IOException {
+    try {
+      return future.join();
+    } catch (CompletionException e) {
+      Throwable cause = e.getCause();
+      if (cause instanceof UncheckedIOException) {
+        throw ((UncheckedIOException) cause).getCause();
+      }
+      throw e;
+    }
   }
 
   /**
--- a/common/src/main/java/software/amazon/s3/analyticsaccelerator/common/telemetry/Telemetry.java
+++ b/common/src/main/java/software/amazon/s3/analyticsaccelerator/common/telemetry/Telemetry.java
@@ -16,6 +16,7 @@
 package software.amazon.s3.analyticsaccelerator.common.telemetry;
 
 import java.io.Closeable;
+import java.io.IOException;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Predicate;
@@ -81,11 +82,13 @@ public interface Telemetry extends Closeable {
    * @param operationSupplier operation to record this execution as.
    * @param operationCode the future to measure the execution of.
    * @return an instance of {@link T} that returns the same result as the one passed in.
+   * @throws IOException if the underlying operation threw an IOException
    */
   default <T> T measureJoin(
       @NonNull TelemetryLevel level,
       @NonNull OperationSupplier operationSupplier,
-      @NonNull CompletableFuture<T> operationCode) {
+      @NonNull CompletableFuture<T> operationCode)
+      throws IOException {
     if (operationCode.isDone()) {
       return operationCode.join();
     } else {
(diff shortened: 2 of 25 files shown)


common/build.gradle.kts
```diff
@@ -10,11 +10,13 @@ plugins {
 
 dependencies {
     implementation(libs.slf4j.api)
+    implementation(libs.s3)
```
CsengerG, line 13: What motivates the new dependency? Is there any way to avoid it?

common/src/main/java/software/amazon/s3/analyticsaccelerator/common/telemetry/Telemetry.java
```diff
@@ -81,11 +82,13 @@ <T> CompletableFuture<T> measure(
...
    * @param operationCode the future to measure the execution of.
    * @return an instance of {@link T} that returns the same result as the one passed in.
+   * @throws IOException if the underlying operation threw an IOException
    */
   default <T> T measureJoin(
       @NonNull TelemetryLevel level,
       @NonNull OperationSupplier operationSupplier,
-      @NonNull CompletableFuture<T> operationCode) {
+      @NonNull CompletableFuture<T> operationCode)
+      throws IOException {
```
CsengerG, line 91: How does this work? Do we still get telemetry logs for something that errored out or is the measurement process interrupted?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Blob.java
```diff
@@ -133,12 +135,10 @@ public IOPlanExecution execute(IOPlan plan) {
                 .build(),
         () -> {
           try {
-            plan.getPrefetchRanges()
-                .forEach(
-                    range -> {
-                      this.blockManager.makeRangeAvailable(
-                          range.getStart(), range.getLength(), ReadMode.ASYNC);
-                    });
+            for (Range range : plan.getPrefetchRanges()) {
```
fuatbasik, line 138: just out of curiosity, why you wanted to change the syntax?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockStore.java
```diff
@@ -88,7 +89,7 @@ public OptionalLong findNextLoadedByte(long pos) {
    * @return the position of the next byte NOT present in the BlockStore or empty if all bytes are
    *     present
    */
-  public OptionalLong findNextMissingByte(long pos) {
+  public OptionalLong findNextMissingByte(long pos) throws IOException {
```
fuatbasik, line 92: Which method here actually throws? Should they throw?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/IOPlanner.java
```diff
@@ -47,7 +48,7 @@ public IOPlanner(@NonNull BlockStore blockStore) {
    * @param lastObjectByte the zero-indexed position of the last object byte
    * @return a list of Ranges that need to be fetched
    */
-  public List<Range> planRead(long pos, long end, long lastObjectByte) {
+  public List<Range> planRead(long pos, long end, long lastObjectByte) throws IOException {
```
fuatbasik, line 51: if this is only Planning a read and not doing it, should it throw?

fuatbasik (changes requested):
Thanks a lot @SanjayMarreddi . 

I put some comments. 2 big ones 1/I think we should move ObjectClient specific exceptionHandling to ObjectClient and refrain from adding an S3 dependency to common. 2/Looks like some methods shouldn't be throwing are throwing today (at least it is not immediately visible to me why they should throw). This might indicate one of the 2 things, i/we are adding a throws statement when it is not needed ii/these methods are actually going beyond their concern and actually perform IO. I put some comments to the obvious ones.

common/build.gradle.kts
SanjayMarreddi: The `ExceptionHandler` class that's under this `common` directory needs this for using the imports from `software.amazon.awssdk.*`.  We can avoid this by moving this class to the `object-client`. ( as also suggested in the below comments ). I'll do that.

common/src/main/java/software/amazon/s3/analyticsaccelerator/common/telemetry/DefaultTelemetry.java
```diff
@@ -344,6 +395,8 @@ private static void recordDatapoint(
     }
   }
 
+  // Put that her
```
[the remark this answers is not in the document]
SanjayMarreddi: Oh sorry, Forgot to remove my self-notes. will do.

common/src/main/java/software/amazon/s3/analyticsaccelerator/common/exceptions/ExceptionHandler.java
```diff
@@ -0,0 +1,85 @@
...
+import java.io.IOException;
+import java.util.stream.Stream;
+import software.amazon.awssdk.core.exception.SdkClientException;
+import software.amazon.awssdk.core.exception.SdkException;
+import software.amazon.awssdk.services.s3.model.InvalidObjectStateException;
+import software.amazon.awssdk.services.s3.model.NoSuchKeyException;
+import software.amazon.awssdk.services.s3.model.S3Exception;
+import software.amazon.s3.analyticsaccelerator.util.S3URI;
+
+/** Handles mapping of S3 exceptions to IO exceptions. */
```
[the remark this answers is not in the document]
SanjayMarreddi: Agree, will move it

common/build.gradle.kts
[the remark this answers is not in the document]
SanjayMarreddi: Sure, I'll move `ExceptionHandler` to `object-client` directory. So we no longer need this.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Blob.java
SanjayMarreddi: The `forEach` lambda doesn't allow throwing checked exceptions (like `IOException` from `makeRangeAvailable`). We'd have to wrap them in `RuntimeException` there.
Regular for loops don't have this limitation, so I thought its cleaner for handling IOExceptions.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockStore.java
SanjayMarreddi: So `getLastObjectByte()` method in this method throws as it gets data using `this.metadataStore.get(s3URI)` which internally does the `headObject` call.  So based on the current way code is structured, it makes me feel its fine for it to throw. Please let me know your thoughts if you disagree.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/IOPlanner.java
SanjayMarreddi: similar to above, this depends upon the `findNextMissingByte` and goes on to make the `headObject` call which can internally throw. ( though we are not reading using `getObject`. )

object-client/src/main/java/software/amazon/s3/analyticsaccelerator/S3SdkObjectClient.java
```diff
@@ -193,7 +201,20 @@ public CompletableFuture<ObjectContent> getObject(
...
+  private <T> Function<Throwable, T> handleException(S3URI s3Uri) {
+    return throwable -> {
+      Throwable cause =
+          Optional.ofNullable(throwable.getCause())
+              .filter(
+                  t ->
+                      throwable instanceof CompletionException
+                          || throwable instanceof ExecutionException)
+              .orElse(throwable);
+      throw new UncheckedIOException(ExceptionHandler.toIOException(cause, s3Uri));
```
fuatbasik, line 217: reading this again, i got confused why are we wrapping the IOException with UncheckedIOException. Why do we need it?

SanjayMarreddi: So, my thought process is this:
 
 - `handleException` method is used in a `CompletableFuture` chain, specifically in the `exceptionally` method above. Since `exceptionally` is a functional style, it doesn't allow the direct throwing of checked exceptions like `IOException.` 
 - By wrapping it in `UncheckedIOException,` we can propagate the exception in a way compatible with the functional-style code while deferring handling to higher-level logic. 
 - I felt this is cleaner than alternatives such as throwing as `RuntimeException` or moving to imperative style using try-catch.
 
Please let me know if you have any suggestions, and correct me if I'm wrong.

CsengerG (approved):

PR #211, opened 2025-01-21
Improve the unit and integration tests

SanjayMarreddi:
## Description of change
- Ensure integration tests run for both S3 Client Kinds
- Add more unit tests for ObjectClient and S3SeekableInputStreamFactory

NO

NO

Unit and Integration tests

NO

---


stubz151:
Do we want to run all tests twice? we already have some big data, and these integ tests aren't fast. Can we not just make half run with one client and half with the other or something like that

SanjayMarreddi:
> Do we want to run all tests twice? we already have some big data, and these integ tests aren't fast. Can we not just make half run with one client and half with the other or something like that

From my high level understanding, yes, we do need to run all tests twice - with both the client kinds as technically, users can use either of them for their use cases. We want to be confident in both the paths. But yeah, may be we can adjust/modify the existing integration tests to make them run quick ( not sure what if we want to tradeoff that way ) 

@fuatbasik Do you have any additional inputs on this?

stubz151 (approved):

PR #216, opened 2025-02-05
Explicitly close input stream.

ahmarsuhail:
## Description of change

After reading from the input stream, we were not explicitly calling close() on it. 

#### Relevant issues

N/A

No

No

Ran tests locally

---


input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/util/StreamUtils.java
```diff
@@ -43,6 +43,8 @@ public static byte[] toByteArray(ObjectContent objectContent) {
       while ((numBytesRead = inStream.read(buffer, 0, buffer.length)) != -1) {
         outStream.write(buffer, 0, numBytesRead);
       }
+
+      inStream.close();
```
fuatbasik, line 47: should we move this to finally block so we are not letting streams around if we catch an exception?

fuatbasik (changes requested):

PR #223, opened 2025-02-12
Adds in constructor for open file information when creating stream.

ahmarsuhail:
## Description of change

This PR adds in a new constructor `public S3SeekableInputStream createStream(
      @NonNull S3URI s3URI, @NonNull OpenFileInformation openFileInformation) `

This can be used to pass in the streamContext, ObjectMetadata, inputPolicy and whatever else is needed later. 

We need to pass in the input policy, as applications can use a `fs.s3a.experimental.input.fadvise` config 
when opening files with S3A. There are cases when we must respect this value, eg for DISTCP, if fadvise is set to `whole-file`, then we shouldn't use parquet optimisations, even if the file format is parquet. 

#### Relevant issues

N/A

No

Yes, adds in new public constructor.

Ran unit tests locally. 

---

### DIFF
--- a/common/src/main/java/software/amazon/s3/analyticsaccelerator/util/InputPolicy.java
+++ b/common/src/main/java/software/amazon/s3/analyticsaccelerator/util/InputPolicy.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package software.amazon.s3.analyticsaccelerator.util;
+
+/**
+ * Input policy to be used when reading a while. Useful for integrating applications to pass down a
+ * policy they would like AAL to use, rather than us guessing based on file format.
+ */
+public enum InputPolicy {
+  None,
+  Sequential
+}
--- a/common/src/main/java/software/amazon/s3/analyticsaccelerator/util/InputPolicy.java
+++ b/common/src/main/java/software/amazon/s3/analyticsaccelerator/util/InputPolicy.java
@@ -16,8 +16,15 @@
 package software.amazon.s3.analyticsaccelerator.util;
 
 /**
- * Input policy to be used when reading a while. Useful for integrating applications to pass down a
+ * Input policy to be used when reading a file. Useful for integrating applications to pass down a
  * policy they would like AAL to use, rather than us guessing based on file format.
+ *
+ * <p>A sequential policy means that the file will be read sequentially, regardless of the file
+ * format. Useful for sequential data types such as CSV, or for applications that read columnar
+ * formats sequentially.
+ *
+ * <p>When no input policy is supplied, AAL will infer what optimisations to use based on the file
+ * extension.
  */
 public enum InputPolicy {
   None,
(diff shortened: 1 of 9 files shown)


input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStreamFactory.java
```diff
@@ -96,36 +96,40 @@ public S3SeekableInputStream createStream(@NonNull S3URI s3URI) throws IOExcepti
...
-
-    objectMetadataStore.storeObjectMetadata(s3URI, metadata);
-
+    storeObjectMetadata(s3URI, metadata);
     return new S3SeekableInputStream(s3URI, createLogicalIO(s3URI), telemetry);
   }
 
   /**
-   * Create an instance of S3SeekableInputStream with streamContext.
+   * Creates and instance of SeekableStream with file information
```
stubz151, line 104: creates and

common/src/main/java/software/amazon/s3/analyticsaccelerator/util/InputPolicy.java
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package software.amazon.s3.analyticsaccelerator.util;
+
+/**
+ * Input policy to be used when reading a while. Useful for integrating applications to pass down a
```
rajdchak, line 19: typo: while -> file

common/src/main/java/software/amazon/s3/analyticsaccelerator/util/InputPolicy.java
```diff
@@ -0,0 +1,25 @@
...
+ * policy they would like AAL to use, rather than us guessing based on file format.
+ */
+public enum InputPolicy {
+  None,
+  Sequential
+}
```
rajdchak, line 25: Can we describe what these policies mean in the javadocs?

rajdchak: From what I understand these input policies are specific to S3A integration, should we add S3A prefix to these policies? Or have separate Input policy enum for each integration. Not sure what is the best way.

ahmarsuhail: not necessarily related to S3A, can be used by any integration application to specify how they will be reading the data. For eg, if S3FileIO is reading CSV it can specify that using this input policy rather than us having to infer from the file extension

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/util/ObjectFormatSelector.java
```diff
@@ -38,9 +38,22 @@ public ObjectFormatSelector(LogicalIOConfiguration configuration) {
    * Uses a regex matcher to select the file format based on the file extension of the key.
    *
    * @param s3URI the object's S3 URI
+   * @param openFileInformation known file information for the fil e
```
stubz151, line 41: [nit] fil e

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/MetadataStore.java
```diff
@@ -128,7 +128,9 @@ public synchronized CompletableFuture<ObjectMetadata> asyncGet(S3URI s3URI) {
    * @param objectMetadata Object metadata
    */
   public synchronized void storeObjectMetadata(S3URI s3URI, ObjectMetadata objectMetadata) {
-    this.cache.put(s3URI, CompletableFuture.completedFuture(objectMetadata));
+    if (objectMetadata != null) {
```
stubz151, line 131: [nit] I don't know if we want to put that if null logic in here, could lead to some misses down the road where we think we are putting metadata but we actually passing a null. Could we move it the factory?

stubz151 (approved):

rajdchak (dismissed):

stubz151 (approved):

rajdchak (approved):

PR #225, opened 2025-02-14
Disable sequential prefetching for non Parquet objects

vaibhav5140:
## Description of change

Added support for selective sequential prefetching based on file type. This change enables sequential prefetching specifically for Parquet files and disabling it for non Parquet

#### Does this contribution introduce any breaking changes to the existing APIs or behaviours?
Yes, Disabling sequential prefetching for non Parquet objects

#### Does this contribution introduce any new public APIs or behaviours?
No

Updated unit tests for BlockManager, PhysicalIOImpl and S3SeekableInputStreamFactory
Verified behaviour with Parquet and non-Parquet files
Modified existing tests to handle OpenFileInformation

No

### DIFF
--- a/common/src/main/java/software/amazon/s3/analyticsaccelerator/util/OpenFileInformation.java
+++ b/common/src/main/java/software/amazon/s3/analyticsaccelerator/util/OpenFileInformation.java
@@ -25,7 +25,7 @@ import software.amazon.s3.analyticsaccelerator.request.StreamContext;
  * information and callbacks when opening the file.
  */
 @Value
-@Builder
+@Builder(toBuilder = true)
 public class OpenFileInformation {
   StreamContext streamContext;
   ObjectMetadata objectMetadata;
--- a/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStreamFactory.java
+++ b/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStreamFactory.java
@@ -28,6 +28,7 @@ import software.amazon.s3.analyticsaccelerator.io.physical.data.MetadataStore;
 import software.amazon.s3.analyticsaccelerator.io.physical.impl.PhysicalIOImpl;
 import software.amazon.s3.analyticsaccelerator.request.ObjectClient;
 import software.amazon.s3.analyticsaccelerator.request.ObjectMetadata;
+import software.amazon.s3.analyticsaccelerator.util.InputPolicy;
 import software.amazon.s3.analyticsaccelerator.util.ObjectFormatSelector;
 import software.amazon.s3.analyticsaccelerator.util.OpenFileInformation;
 import software.amazon.s3.analyticsaccelerator.util.S3URI;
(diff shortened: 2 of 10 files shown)


ahmarsuhail (changes requested):
looks good overall, 

i'm not sure if block store tests cases are quite right, or maybe i misunderstood them. let's discuss

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStore.java
```diff
@@ -69,17 +70,22 @@ protected boolean removeEldestEntry(final Map.Entry<ObjectKey, Blob> eldest) {
...
    * @param metadata the metadata for the object we are computing
    * @param streamContext contains audit headers to be attached in the request header
+   * @param openFileInformation known file information including input policy
    * @return the blob representing the object from the BlobStore
    */
-  public Blob get(ObjectKey objectKey, ObjectMetadata metadata, StreamContext streamContext) {
+  public Blob get(
+      ObjectKey objectKey,
+      ObjectMetadata metadata,
+      StreamContext streamContext,
```
SanjayMarreddi, line 79: `streamContext` This is no longer being passed to BlockManager right? Should we remove it?

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManagerTest.java
```diff
@@ -286,6 +294,115 @@ private BlockManager getTestBlockManager(
...
+    verify(objectClient, atLeast(1)).getObject(requestCaptor.capture(), any());
+
+    // Verify requests use sequential prefetching
+    requestCaptor
+        .getAllValues()
+        .forEach(
+            request ->
+                assertTrue(
+                    request.getRange().getLength() <= testConfig.getReadAheadBytes(),
+                    "Parquet requests should show sequential prefetching"));
```
SanjayMarreddi, line 406: I think we should move the common code between these tests into separate functions - so that tests will be easy to follow

common/src/main/java/software/amazon/s3/analyticsaccelerator/util/OpenFileInformation.java
```diff
@@ -25,7 +25,7 @@
  * information and callbacks when opening the file.
  */
 @Value
-@Builder
+@Builder(toBuilder = true)
```
stubz151, line 28: what does this do? and why are we doing it?

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStreamFactoryTest.java
```diff
@@ -17,8 +17,7 @@
 
 import static org.junit.jupiter.api.Assertions.*;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
stubz151, line 20: don't use * imports turn this off in intellij
https://stackoverflow.com/questions/3348816/intellij-never-use-wildcard-imports

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStreamFactory.java
```diff
@@ -117,31 +118,27 @@ LogicalIO createLogicalIO(S3URI s3URI) throws IOException {
...
-                objectBlobStore,
-                telemetry,
-                openFileInformation.getStreamContext()),
+                s3URI, objectMetadataStore, objectBlobStore, telemetry, openFileInformation),
             telemetry,
             configuration.getLogicalIOConfiguration(),
             parquetColumnPrefetchStore);
 
       default:
+        OpenFileInformation effectiveInfo = openFileInformation;
```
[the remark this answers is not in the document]
rajdchak: why do we need to create a new OpenFileInformation object again, can we not set the input policy in the existing object itself?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/impl/PhysicalIOImpl.java
```diff
@@ -71,19 +72,20 @@ public PhysicalIOImpl(
...
-   * @param streamContext contains audit headers to be attached in the request header
+   * @param openFileInformation known file information including input policy
    */
   public PhysicalIOImpl(
       @NonNull S3URI s3URI,
       @NonNull MetadataStore metadataStore,
       @NonNull BlobStore blobStore,
       @NonNull Telemetry telemetry,
-      StreamContext streamContext)
+      OpenFileInformation openFileInformation)
```
rajdchak, line 82: add the @NotNull check

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/impl/PhysicalIOImpl.java
```diff
@@ -71,19 +72,20 @@ public PhysicalIOImpl(
...
       throws IOException {
     this.metadataStore = metadataStore;
     this.blobStore = blobStore;
     this.telemetry = telemetry;
-    this.streamContext = streamContext;
+    this.openFileInformation = openFileInformation;
+    this.streamContext = openFileInformation.getStreamContext();
```
rajdchak, line 88: we can remove streamContext from here, only pass openFileInformation to blobStore which will have the streamContext.

PR #238, opened 2025-03-04
Optimise Sequential Format Reader

vaibhav5140:
## Description of change

Introduce Optimised Reading strategy for sequential access pattern formats (CSV,  JSON, TXT)

- Implements a prefetching strategy that prefetches up to the partition size(Async prefetch)
- Introduces a new configuration parameter, which by default is set to 128MB
- Optimized for Spark's chunk reading pattern, aligning with the default 128MB partition size.
- Extends DefaultLogicalIOImpl with a new SequentialLogicalIOImpl
- Updates S3SeekableInputStreamFactory to use new implementation

Yes, this contribution introduces a change in behaviour for files with InputPolicy=Sequential:
- Previously, files with InputPolicy=Sequential were handled by DefaultLogicalIOImpl.
- Now, these files are handled by the new SequentialLogicalIOImpl.

Yes, The new approach is designed to work effectively for all sequential access pattern formats.


vaibhav5140:
I am opening this as a draft PR to get feedback on the logical implementation of Non-parquet Reader Optimisation
NOTE: Testing and full integration are still pending. This review is focused on the logical structure and approach.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/impl/OptimizedLogicalIOImpl.java
```diff
@@ -0,0 +1,145 @@
...
+    long contentLength = metadata().getContentLength(); // Get the total length of the file
+    long remainingLength =
+        contentLength - initialPosition; // Calculate how much of the file is left to read
+
+    // First, read the initial chunk,Determine how much to read initially (64KB or less if file is
+    // smaller)
+    long initialReadSize = Math.min(INITIAL_READ_SIZE, remainingLength);
+    if (initialReadSize > 0) {
+      try {
+        getPhysicalIO()
```
ahmarsuhail, line 71: you want to use physicalIO.execute()

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/impl/OptimizedLogicalIOImpl.java
```diff
@@ -0,0 +1,145 @@
...
+      @NonNull S3URI s3URI, @NonNull PhysicalIO physicalIO, @NonNull Telemetry telemetry) {
+    super(s3URI, physicalIO, telemetry);
+  }
+
+  @Override
+  // Overrides the read method from the superclass i.e. DefaultLogicalIOImpl
+  public int read(byte[] buf, int off, int len, long position) throws IOException {
+
+    if (!prefetchStarted) { // If prefetching hasn't started, initiate it
+      startPrefetch(position);
```
ahmarsuhail, line 49: why wait for the first read for a prefetch?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/impl/OptimizedLogicalIOImpl.java
```diff
@@ -0,0 +1,145 @@
...
+  // Overrides the close method from the superclass
+  public void close() throws IOException {
+    if (prefetchFuture != null) {
+      prefetchFuture.cancel(true);
+    }
+    super.close();
+  }
+}
+/*
+APPROACH:
```
rajdchak, line 123: I have some questions about the current implementation:

1. Why did we choose this specific approach for prefetching?
2. Is this approach optimal for all our expected file sizes?
3. Can we add some benchmarks or performance metrics to justify this implementation?

vaibhav5140: 1. Reason for the chosen prefetching approach:
   After exploring various approaches, we found this implementation to be the most optimal. Key reasons:
   - We prefetch data up to 128MB, aligning with Spark's chunk reading pattern.
   - The initial small, direct request ensures the reader doesn't wait.
   - Simultaneous prefetching of the rest of the file (up to 128MB) prepares data for subsequent reads.
   - Our existing range optimiser at the physical layer splits the larger prefetch chunk into 8MB requests, ensuring efficient data retrieval.
   - For sequential formats like CSV, where we ultimately need all the data, this approach provides a good balance between immediate access and preparing subsequent data.

2. Optimality for different file sizes:
   This approach is to be optimal across various file sizes:
   - For files smaller than 128MB, we prefetch the entire file.
   - For larger files, we prefetch up to 128MB, which aligns with Spark's reading pattern.
   This strategy ensures we're not over fetching for small files while providing optimal performance for larger files.

3. Benchmark justification:
   We have conducted tests with this implementation:
   - TPCDS benchmark (1 iteration, 104 queries) showed a 3% improvement in overall execution time.

4. Implementation approach:
   While we initially tested by modifying the PhysicalIO configurations and prefetching logic at the physical layer, we recognized that the correct approach is to implement this at the logical layer. This new OptimizedLogicalIOImpl provides a non-Parquet specific implementation at the logical level, which is more maintainable.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/impl/CSVPrefetcher.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+  }
+  /**
+   * Starts the prefetch process. This includes an initial synchronous read followed by an
+   * asynchronous prefetch of a larger chunk.
+   *
+   * @param initialPosition the position to start prefetching from
+   * @throws IOException if an I/O error occurs
+   */
+  private void startPrefetch(long initialPosition) throws IOException {
```
ahmarsuhail, line 84: you can cut mostly all of this. there is no need for a future here, your underlying operations are async already. 

So all you need to do is 

if (prefetchStarted) {
      return;
    }
    prefetchStarted = true;

PhyiscalIo.execute(currentPos-currentPos + 128MB)

as discussed, drop the splitting of ranges. You actually don't need the configs then too

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/impl/CSVPrefetcher.java
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * A prefetcher optimizing read performance for sequential file formats, currently focused on CSV.
+ * It performs a small initial read followed by an asynchronous larger prefetch.
+ *
+ * <p>Configurable read sizes allow optimization for various use cases. While designed for CSV, this
+ * approach is applicable to other sequentially read formats, potentially improving performance by
+ * reducing individual read requests and leveraging asynchronous prefetching.
+ */
+public class CSVPrefetcher implements AutoCloseable {
```
ahmarsuhail, line 39: discussed offline, rename to SequentialPrefetcher

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/LogicalIOConfiguration.java
```diff
@@ -43,6 +43,9 @@ public class LogicalIOConfiguration {
   private static final int DEFAULT_MAX_COLUMN_ACCESS_STORE_SIZE = 15;
   private static final String DEFAULT_PARQUET_FORMAT_SELECTOR_REGEX = "^.*.(parquet|par)$";
   private static final PrefetchMode DEFAULT_PREFETCHING_MODE = PrefetchMode.ROW_GROUP;
+  private static final String DEFAULT_CSV_FORMAT_SELECTOR_REGEX = "^.*\\.(csv|CSV)$";
+  private static final long DEFAULT_CSV_INITIAL_READ_SIZE = 2 * ONE_MB;
```
ahmarsuhail, line 47: can get rid off  both `DEFAULT_CSV_INITIAL_READ_SIZE` and `DEFAULT_CSV_PREFETCH_SIZE`, instead add a new config 

`SPARK_PARTITION_SIZE`.. , with default as 128MB. Let's users set partition size, will be useful for parquet too! lmk if you agree @fuatbasik

ahmarsuhail (changes requested):
thanks, good start!

haven't reviewed test code yet.. left some comments. 

Mainly, we must make sure we remove the key from the block store on the close. We can't do that for Parquet, it's complicated there. but you should do this for CSV

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/impl/CSVPrefetcher.java
```diff
@@ -0,0 +1,141 @@
...
+  private void startPrefetch(long initialPosition) throws IOException {
+    if (prefetchStarted) {
+      return;
+    }
+    prefetchStarted = true;
+
+    long contentLength = physicalIO.metadata().getContentLength();
+    long remainingLength = contentLength - initialPosition;
+
+    long initialReadSize = Math.min(this.initialReadSize, remainingLength);
```
stubz151, line 93: You shouldn't use a variable name that is already scoped in the class can be weird for everyone else, also is initialReadSize the correct name here? should it not just be readSize?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/impl/CSVPrefetcher.java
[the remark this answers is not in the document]
fuatbasik: who should make decision to how much to prefetch. LogicalIO or Prefetcher? If LogicalIO is the one to make it can we actually re-use it to implement sequential reads of entire objects even when object is not sequential?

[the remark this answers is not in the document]
fuatbasik: we need to add telemetry to prefetcher. See ParquetPrefetcher.

fuatbasik (changes requested):
Thanks @vaibhav5140 . I put some comments. 

Overall, i think we should move away from using the term CSV all together and mention Sequential read patterns. Also, there are some responsibilities issues i raised, it is not clear to me which component should be responsible from what.

Also, i like to understand how can we generalize SequentialPrefetcher to more use-cases. Is there a way for ParquetPrefetcher and SequentialPrefetcher (you are calling it CSV) to share a common interface? How does this change interact with InputPolicy.sequential?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/impl/CSVPrefetcher.java
```diff
@@ -0,0 +1,141 @@
...
+   * asynchronous prefetch of a larger chunk.
+   *
+   * @param initialPosition the position to start prefetching from
+   * @throws IOException if an I/O error occurs
+   */
+  private void startPrefetch(long initialPosition) throws IOException {
+    if (prefetchStarted) {
+      return;
+    }
+    prefetchStarted = true;
```
stubz151, line 88: A bit confused on this logic when do we ever set it to false? If we just doing this to ensure only one instance of the prefetcher is running at once, think we can start it as a task and just check that the task is running. Also if we do do that, this class shouldn't be in charge of that, it should be the caller class or something like that.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/impl/CSVPrefetcher.java
```diff
@@ -0,0 +1,141 @@
...
+
+    long contentLength = physicalIO.metadata().getContentLength();
+    long remainingLength = contentLength - initialPosition;
+
+    long initialReadSize = Math.min(this.initialReadSize, remainingLength);
+    IOPlan initialPlan =
+        new IOPlan(new Range(initialPosition, initialPosition + initialReadSize - 1));
+    IOPlanExecution initialExecution = physicalIO.execute(initialPlan);
+
+    if (initialExecution.getState() != IOPlanState.SUBMITTED) {
```
stubz151, line 98: Can you explain this? I don't really get what it's doing.

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/util/ObjectFormatSelectorTest.java
```diff
@@ -50,7 +50,32 @@ public void testConfiguredExtensionParquetLogicalIOSelection(String key) {
...
+            LogicalIOConfiguration.builder().csvFormatSelectorRegex("^.*.(data|csvdata)$").build());
+
+    assertEquals(
+        ObjectFormat.CSV,
+        objectFormatSelector.getObjectFormat(
+            S3URI.of("bucket", key), OpenStreamInformation.DEFAULT));
+  }
+
+  @ParameterizedTest
+  @ValueSource(strings = {"key.jar", "key.txt", "key.parque", "key.pa", "key.cs"})
```
stubz151, line 78: can we add some cases with csv/parquet in the names? just to validate we looking at the extension correctly.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/util/ObjectFormatSelector.java
```diff
@@ -51,7 +51,7 @@ public ObjectFormat getObjectFormat(S3URI s3URI, OpenStreamInformation openStrea
     // parquet specific optimisations are of no use there :(
     if (openStreamInformation.getInputPolicy() != null
         && openStreamInformation.getInputPolicy().equals(InputPolicy.Sequential)) {
-      return ObjectFormat.DEFAULT;
+      return ObjectFormat.SEQUENTIAL;
```
ahmarsuhail, line 54: Currently you are only using this if input policy is sequential.

you also need to use it if file format is <CSV, JSON, TXT>. Otherwise use default.

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/io/logical/impl/SequentialLogicalIOImplTest.java
```diff
@@ -0,0 +1,121 @@
...
+                null, physicalIO, TestTelemetry.DEFAULT, config, blobStore));
+
+    assertThrows(
+        NullPointerException.class,
+        () ->
+            new SequentialLogicalIOImpl(TEST_URI, physicalIO, TestTelemetry.DEFAULT, config, null));
+  }
+
+  @Test
+  void testReadWithPrefetch() throws IOException {
```
ahmarsuhail, line 84: this is the wrong test for this class. what you need to do here is make sure your prefetcher is called on the read.

For that you need to a new constructor that does dependency injection of your prefetcher, and you want assert that your prefetcher was called.

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/io/logical/impl/SequentialLogicalIOImplTest.java
```diff
@@ -0,0 +1,121 @@
...
+    when(mockMetadata.getEtag()).thenReturn("testEtag");
+    when(physicalIO.metadata()).thenReturn(mockMetadata);
+
+    SequentialLogicalIOImpl logicalIO =
+        new SequentialLogicalIOImpl(
+            TEST_URI, physicalIO, TestTelemetry.DEFAULT, configuration, blobStore);
+
+    logicalIO.close();
+
+    verify(blobStore).evictKey(any(ObjectKey.class));
```
ahmarsuhail, line 119: cut, as you don't want blob store evictions happening here. assert that physicalIO.close() was called.

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/io/logical/impl/SequentialPrefetcherTest.java
```diff
@@ -0,0 +1,128 @@
...
+    value = "NP_NONNULL_PARAM_VIOLATION",
+    justification = "We mean to pass nulls to checks")
+public class SequentialPrefetcherTest {
+  private static final S3URI TEST_URI = S3URI.of("foo", "bar.data");
+  private PhysicalIO physicalIO;
+  private LogicalIOConfiguration config;
+  private ObjectMetadata metadata;
+
+  @BeforeEach
+  void setUp() {
```
ahmarsuhail, line 43: you don't need this

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/io/logical/impl/SequentialPrefetcherTest.java
```diff
@@ -0,0 +1,128 @@
...
+    when(physicalIO.metadata()).thenReturn(metadata);
+    IOPlanExecution mockExecution = mock(IOPlanExecution.class);
+    when(physicalIO.execute(any(IOPlan.class))).thenReturn(mockExecution);
+
+    SequentialPrefetcher prefetcher =
+        new SequentialPrefetcher(TEST_URI, physicalIO, TestTelemetry.DEFAULT, config);
+
+    // Initial prefetch from start
+    prefetcher.prefetch(0);
+    verify(physicalIO)
```
ahmarsuhail, line 86: use argumentCaptor, see ParquetPredictivePrefetchingTaskTest for an example

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/io/logical/impl/SequentialPrefetcherTest.java
```diff
@@ -0,0 +1,128 @@
...
+                    plan.getPrefetchRanges().get(0).getStart() == 0
+                        && plan.getPrefetchRanges().get(0).getEnd() == 4095));
+
+    // Subsequent prefetches should be ignored
+    prefetcher.prefetch(100);
+    prefetcher.prefetch(200);
+    verify(physicalIO, times(1)).execute(any(IOPlan.class));
+
+    // Prefetch near end of file
+    reset(physicalIO);
```
ahmarsuhail, line 99: what's happening here? don't get what you're testing here

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/io/logical/impl/SequentialPrefetcherTest.java
```diff
@@ -0,0 +1,128 @@
...
+    when(metadata.getContentLength()).thenReturn(3000L);
+    when(physicalIO.metadata()).thenReturn(metadata);
+    when(physicalIO.execute(any(IOPlan.class))).thenReturn(mockExecution);
+
+    SequentialPrefetcher prefetcherSmallFile =
+        new SequentialPrefetcher(TEST_URI, physicalIO, TestTelemetry.DEFAULT, config);
+
+    prefetcherSmallFile.prefetch(2000);
```
ahmarsuhail, line 107: this should be a separate test

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/util/ObjectFormatSelectorTest.java
```diff
@@ -71,6 +71,6 @@ public void testDefaultLogicalIOSelectionWithSequentialInputPolicy(String key) {
         objectFormatSelector.getObjectFormat(
             S3URI.of("bucket", key),
             OpenStreamInformation.builder().inputPolicy(InputPolicy.Sequential).build()),
-        ObjectFormat.DEFAULT);
+        ObjectFormat.SEQUENTIAL);
```
ahmarsuhail, line 74: you need new tests here, for CSV, TXT, JSON

they should all return SEQUENTIAL, even if InputPolicy is not sequential

ahmarsuhail (changes requested):
nearly there...

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/io/logical/impl/SequentialPrefetcherTest.java
vaibhav5140: In this part of test(Prefetch near end of file), goal was to test prefetch range should be till end of file, if end of file<spark_partion_size(i.e. 128MB)

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/io/logical/impl/SequentialLogicalIOImplTest.java
vaibhav5140: I have updated this test in latest commit but I don't want to modify the source code by adding a new constructor for dependency injection. Instead, I am using Mockito's `mockConstruction` feature to intercept the creation of `SequentialPrefetcher` within the `SequentialLogicalIOImpl` constructor. This allows us to test the internal behavior of `SequentialLogicalIOImpl` , verifying that the `prefetch` method is called on the internal `SequentialPrefetcher` when `read` is invoked.

fuatbasik (dismissed):
Thanks @vaibhav5140  I think this needs more work. 
1/ We need integration tests. 
2/ We need a prefetcher interface to have a common interface on multiple prefetching implementations. 
3/ We should never mention Spark in any-place in this code-base. We need to change the configuration flags.

input-stream/src/testFixtures/java/software/amazon/s3/analyticsaccelerator/access/S3Object.java
```diff
@@ -135,4 +142,19 @@ public static List<S3Object> largeObjects() {
   public static List<S3Object> allObjects() {
     return filter(o -> true);
   }
+
+  /**
+   * Returns list of sequential objects (CSV, JSON, and TXT files).
+   *
+   * @return list of sequential objects
+   */
+  public static List<S3Object> getSequentialS3Objects() {
```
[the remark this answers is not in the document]
ahmarsuhail: look at the `smallAndMediumObjects` filter for example

ahmarsuhail (changes requested):
looks ok, some minor changes

common/src/main/java/software/amazon/s3/analyticsaccelerator/common/ConnectorConfiguration.java
```diff
@@ -264,4 +264,22 @@ private String expandKey(String key) {
...
+   * Get Long value for a given key, ensuring the value is positive. If key is not found, return
+   * default value. Note that this method will throw an exception if the value is not positive.
+   *
+   * @param key suffix of the configuration to retrieve. Full search key will be this.getPrefix +
+   *     "." + key
+   * @param defaultValue default value if provided key does not exist
+   * @return long
+   * @throws IllegalArgumentException if the value is not positive
+   */
+  public long getPositiveLong(@NonNull String key, long defaultValue) {
```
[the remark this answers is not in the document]
vaibhav5140: getLong only checks if value can be parsed as a long,
getPositiveLong adds validation that the value must be > 0, which is needed to test that partitionSize should always be positive

input-stream/src/testFixtures/java/software/amazon/s3/analyticsaccelerator/access/S3Object.java
```diff
@@ -135,4 +142,19 @@ public static List<S3Object> largeObjects() {
...
+    return S3Object.filter(
+        obj -> {
+          String lowerName = obj.getName().toLowerCase(Locale.getDefault());
```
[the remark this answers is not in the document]
vaibhav5140: Locale.getDefault() with toLowerCase() because some languages have special case conversion rules, without a Locale, case conversion might produce unexpected results in different languages. I just used it as a good practice to ensure consistent behaviour across all environments.

common/src/main/java/software/amazon/s3/analyticsaccelerator/common/ConnectorConfiguration.java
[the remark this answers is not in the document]
ahmarsuhail: the right way to do this is actually how it's done in `PhysicalIOConfiguration`, the positive validation shouldn't happen here.

Can you create a ticked to add validation to our logicalIoConfiguration , and then we will cut this.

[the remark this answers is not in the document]
ahmarsuhail: you don't need to change anything here for now, dw

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/impl/DefaultLogicalIOImpl.java
```diff
@@ -121,13 +121,16 @@ public ObjectMetadata metadata() throws IOException {
...
+  protected void closeWithEviction(boolean shouldEvict) throws IOException {
+    physicalIO.close(shouldEvict);
+  }
   /**
    * Closes associate resources.
    *
    * @throws IOException
    */
   @Override
   public void close() throws IOException {
```
ahmarsuhail, line 133: can you also create another ticket to track the following:

The default logicalIo should also evict from cache on close, but parquet logicalIO should be modified to not evict.

ahmarsuhail (approved):
+1, LGTM. Pending rebase.

ahmarsuhail (approved):

fuatbasik (commented):
revoking "Requested Changes"

ahmarsuhail (approved):
+1, LGTM

PR #251, opened 2025-04-04
Memory manager changes

rajdchak:
## Description of change
-> Created a caffeine index cache to store indexes of all blocks across all blobs. We index using the BlockKey (added this with the change) which is a combination of the ObjectKey(S3Uri + Etag) and Range(start and end byte).
-> Configured a Caffeine cache with the following properties:
           Entries expire after 1 seconds of inactivity
           Custom weigher to use block size as the weight
           Maximum total weight set to the configured blob store capacity
-> A clean up task to run every 5 seconds and sync the blob map with the index cache.
      

Unit tests and benchmarks. Ran benchmarks with S3A, S3FileIO with 1000mb object size. Seeing almost similar performance with or without memory manager changes. Also metrices published for memory usage show that the memory used has reduced from before.

---


input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStore.java
```diff
@@ -52,17 +58,37 @@ public BlobStore(
...
-            });
+    this.blobMap = Collections.synchronizedMap(new LinkedHashMap<ObjectKey, Blob>());
+    this.indexCache =
+        Caffeine.newBuilder()
+            .expireAfterAccess(5, TimeUnit.SECONDS)
+            .removalListener(this::onRemoval)
+            .weigher((blockId, blockSize) -> blockSize)
+            .maximumWeight(configuration.getBlobStoreCapacity())
+            .build();
     this.configuration = configuration;
```
rajdchak, line 69: // Configure a Caffeine cache with the following properties:
// - Entries expire after 5 seconds of inactivity
// - Custom removal listener to handle evicted entries (blocks with active readers added back, else entry removed from blockstore)
// - Custom weigher to use block size as the weight
// - Maximum total weight set to the configured blob store capacity

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
```diff
@@ -262,7 +269,7 @@ public int read(byte @NonNull [] buf, int off, int len, long pos) throws IOExcep
   public boolean contains(long pos) {
     Preconditions.checkArgument(0 <= pos, "`pos` must not be negative");
 
-    return start <= pos && pos <= end;
+    return this.blockKey.getRange().getStart() <= pos && pos <= this.blockKey.getRange().getEnd();
```
fuatbasik, line 272: should this actually be a method of Range?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStore.java
```diff
@@ -52,17 +58,37 @@ public BlobStore(
...
   }
 
+  private void onRemoval(BlockKey key, Integer value, RemovalCause cause) {
+    Map<BlockKey, Block> blocks =
+        blobMap.get(key.getObjectKey()).getBlockManager().getBlockStore().getBlocks();
```
fuatbasik, line 74: i wonder if we should encapsulate this logic in BlockStore or in BlockManager?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManager.java
```diff
@@ -114,17 +117,19 @@ public synchronized Optional<Block> getBlock(long pos) {
...
+      long pos, ReadMode readMode, Cache<BlockKey, Integer> indexCache) throws IOException {
     Preconditions.checkArgument(0 <= pos, "`pos` must not be negative");
 
     // Position is already available --> return corresponding block
     if (getBlock(pos).isPresent()) {
       return;
     }
 
-    makeRangeAvailable(pos, 1, readMode);
+    makeRangeAvailable(pos, 1, readMode, indexCache);
```
fuatbasik, line 132: indexCache is shared between multiple BlockManagers. Sync. will be useful for the same instance of the BlockManager but we probably need indexCache to be threadSafe too. Otherwise, we can have interesting race-conditions. This is inline with my previous comments about making indexCache a class on its own, implement thread safe methods and call them from the BlockManager or BlockStore.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
```diff
@@ -131,27 +109,34 @@ public Block(
...
-    this.range = new Range(start, end);
+    this.blockKey = blockKey;
     this.objectClient = objectClient;
     this.streamContext = streamContext;
     this.readMode = readMode;
-    this.referrer = new Referrer(range.toHttpString(), readMode);
+    this.referrer = new Referrer(this.blockKey.getRange().toHttpString(), readMode);
     this.readTimeout = readTimeout;
     this.readRetryCount = readRetryCount;
+    this.activeReaders = new AtomicInteger(0);
```
fuatbasik, line 139: now that we have an activeReaders we probably need an isActive method

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStoreTest.java
```diff
@@ -55,7 +61,32 @@ void setUp() throws IOException {
...
+    Map<String, String> configMap = new HashMap<>();
+    configMap.put("blobstore.capacity", "1000");
+    ConnectorConfiguration connectorConfig = new ConnectorConfiguration(configMap);
+    config = PhysicalIOConfiguration.fromConfiguration(connectorConfig);
+
+    blobStore = new BlobStore(objectClient, TestTelemetry.DEFAULT, config);
+  }
+
+  @Test
+  void testIndexCacheEviction() throws InterruptedException {
```
fuatbasik, line 73: if we move IndexCache to its own class, we can have number of test classes. We probably need much more tests

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStore.java
```diff
@@ -52,17 +58,37 @@ public BlobStore(
...
+    Block blockToBeRemoved = blocks.get(key);
+    if (blockToBeRemoved.getActiveReaders().get() > 0) {
+      indexCache.put(key, value);
```
stubz151, line 77: Is there no way to just stop a block being removed if it's busy being read?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
```diff
@@ -30,13 +31,10 @@
...
 import software.amazon.s3.analyticsaccelerator.request.ObjectClient;
 import software.amazon.s3.analyticsaccelerator.request.ObjectContent;
-import software.amazon.s3.analyticsaccelerator.request.Range;
 import software.amazon.s3.analyticsaccelerator.request.ReadMode;
 import software.amazon.s3.analyticsaccelerator.request.Referrer;
 import software.amazon.s3.analyticsaccelerator.request.StreamContext;
-import software.amazon.s3.analyticsaccelerator.util.ObjectKey;
-import software.amazon.s3.analyticsaccelerator.util.StreamAttributes;
-import software.amazon.s3.analyticsaccelerator.util.StreamUtils;
+import software.amazon.s3.analyticsaccelerator.util.*;
```
stubz151, line 37: is this * on purpose?

common/src/main/java/software/amazon/s3/analyticsaccelerator/util/BlockKey.java
```diff
@@ -0,0 +1,31 @@
...
+import lombok.Builder;
+import lombok.Getter;
+import lombok.NonNull;
+import software.amazon.s3.analyticsaccelerator.request.Range;
+
+/** Container used to represent an S3 object for a specific version/etag */
+@Getter
+@AllArgsConstructor
+@Builder
+public class BlockKey {
```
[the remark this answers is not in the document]
stubz151: [Thought] Do we need block key? could we not just expand object key to have this range?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
```diff
@@ -211,6 +199,15 @@ private void generateSourceAndData() throws IOException {
     }
   }
 
+  /**
+   * Updates and returns the active readers of this blob
+   *
+   * @param readers The delta to be added to the current active readers of this blob
+   */
+  public void updateActiveReaders(int readers) {
+    activeReaders.addAndGet(readers);
```
stubz151, line 208: Why addAndGet if we don't use the return?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
```diff
@@ -211,6 +199,15 @@ private void generateSourceAndData() throws IOException {
     }
   }
 
+  /**
+   * Updates and returns the active readers of this blob
```
stubz151, line 203: This method doesn't currently return

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockStore.java
```diff
@@ -17,14 +17,13 @@
 
 import java.io.Closeable;
 import java.io.IOException;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Optional;
-import java.util.OptionalLong;
+import java.util.*;
```
stubz151, line 20: check the * import

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockStore.java
```diff
@@ -61,7 +60,7 @@ public BlockStore(ObjectKey objectKey, ObjectMetadata metadata) {
   public Optional<Block> getBlock(long pos) {
     Preconditions.checkArgument(0 <= pos, "`pos` must not be negative");
 
-    return blocks.stream().filter(b -> b.contains(pos)).findFirst();
+    return blocks.values().stream().filter(b -> b.contains(pos)).findFirst();
```
stubz151, line 63: if we know the range of each block can't we use that to workout the block we need to return here? feel like if we swapping to a datastructure that uses more memory we should try make use of it :D

stubz151 (changes requested):

common/src/main/java/software/amazon/s3/analyticsaccelerator/util/BlockKey.java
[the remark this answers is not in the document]
rajdchak: We cannot include range in object key, because object key is used to identify the blobs whereas the block key will identify the blocks.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Blob.java
```diff
@@ -109,6 +116,7 @@ public int read(byte[] buf, int off, int len, long pos) throws IOException {
                               "This block (for position %s) should have been available.",
                               nextPositionFinal)));
 
+      indexCache.getIfPresent(nextBlock.getBlockKey());
```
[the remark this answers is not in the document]
rajdchak: we do not need the result from index cache, but we since we are using the corresponding data from blob map, we need to make the call on index cache to update the recency of access of the corresponding block key.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockStore.java
rajdchak: may be we can do some optimisations there like keeping the block keys sorted according to ranges and then we could spot the block using binary search or some other efficient algo. But I don't want to do this as part of the memory manager changes. Can take that up as a follow up task.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
```diff
@@ -131,27 +109,34 @@ public Block(
 
     Preconditions.checkArgument(
         0 <= generation, "`generation` must be non-negative; was: %s", generation);
-    Preconditions.checkArgument(0 <= start, "`start` must be non-negative; was: %s", start);
-    Preconditions.checkArgument(0 <= end, "`end` must be non-negative; was: %s", end);
     Preconditions.checkArgument(
-        start <= end, "`start` must be less than `end`; %s is not less than %s", start, end);
+        0 <= blockKey.getRange().getStart(),
```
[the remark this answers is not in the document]
rajdchak: yes will make this change

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStoreIndexCache.java
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  /**
+   * Checks if a specific block key exists in the cache.
+   *
+   * @param blockKey the key to check for existence
+   * @return true if the block key exists in the cache, false otherwise
+   */
+  public boolean contains(BlockKey blockKey) {
+    return indexCache.asMap().containsKey(blockKey);
```
fuatbasik, line 68: nit: is this converting Cache to a map everytime we check contains?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
```diff
@@ -228,6 +223,11 @@ public int read(long pos) throws IOException {
     Preconditions.checkArgument(0 <= pos, "`pos` must not be negative");
 
     byte[] content = this.getDataWithRetries();
+    if (!blockMetricsAndCacheHandler.isPresentInIndexCache(blockKey)) {
```
fuatbasik, line 226: nit: shall we move this to a method in blockMetricsAndCacheHandler or in this class? I am assuming this is to record cache hit/miss right?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
fuatbasik: we can write a contains method to range to tell us if a range contains a given position.

input-stream/build.gradle.kts
```diff
@@ -145,6 +146,7 @@ val shadowJar = tasks.withType<ShadowJar> {
 
     relocate("org.apache.parquet.format", "software.amazon.s3.shaded.apache.parquet.format")
     relocate("shaded.parquet.org.apache.thrift", "software.amazon.s3.shaded.parquet.org.apache.thrift")
+    relocate("com.github.benmanes.caffeine", "com.amazon.shaded.github.benmanes.caffeine")
```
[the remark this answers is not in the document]
rajdchak: ack will update this

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Blob.java
```diff
@@ -71,18 +73,27 @@ public Blob(
...
 
-  /**
-   * Returns the memory used by the blob
-   *
-   * @return the memory used by the blob
-   */
-  public long getMemoryUsageOfBlob() {
+    try {
+      lock.readLock().lock();
+      blockManager.makePositionAvailable(pos, ReadMode.SYNC);
```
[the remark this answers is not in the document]
rajdchak: ack

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStore.java
```diff
@@ -64,28 +71,51 @@ public BlobStore(
...
+        cleanupInProgress.set(false);
+      }
+    }
+  }
+
+  void asyncCleanup() {
+    LOG.debug(
+        "Current memory usage of blobMap in bytes before eviction is: {}",
+        metrics.get(MetricKey.MEMORY_USAGE));
+    blobMap.forEach((k, v) -> v.asyncCleanup());
```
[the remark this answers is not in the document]
rajdchak: no we are not removing the blob object, we are just clearing out the blocks in it because that is the actual data consuming memory. I don't think it is necessary to remove the blob object, we can have a read again on the same object.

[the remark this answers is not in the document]
rajdchak: and as per logic, if the blob object does not have any blocks present the cleanup task doesn't enter to clean it.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStoreIndexCache.java
rajdchak: No, Caffeine's asMap() method does not convert the cache to a map. It provides a thread-safe map view of the cache, and modifications to this view directly affect the cache.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
rajdchak: we are recording cache hit/miss directly from our blob map in the blockstore class. We don't care about the hit/miss of the index cache.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
```diff
@@ -228,6 +223,11 @@ public int read(long pos) throws IOException {
...
+      blockMetricsAndCacheHandler.putInIndexCache(blockKey, blockKey.getRange().getLength());
+    } else {
+      blockMetricsAndCacheHandler.getIfPresentFromIndexCache(blockKey);
```
ahmarsuhail, line 229: is this just to record an access to the key so it doesn't expire?

If yes, can you rename the method so it's clear why we have it

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
ahmarsuhail: +1, pls add a new `contains()` method in the Range class. I know this is unrelated to your change and was always like this, but since you're updating this line might as well fix now :)

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockStore.java
```diff
@@ -121,11 +133,46 @@ public OptionalLong findNextMissingByte(long pos) throws IOException {
...
+   */
+  public void cleanUp() {
+
+    Iterator<Map.Entry<BlockKey, Block>> iterator = blocks.entrySet().iterator();
+
+    while (iterator.hasNext()) {
+      Map.Entry<BlockKey, Block> entry = iterator.next();
+      BlockKey blockKey = entry.getKey();
+
+      if (entry.getValue().isDataLoaded()
```
fuatbasik, line 158: So this is saying if data is not loaded yet we will not remove that block, even if our eviction algorithm tells us to remove it right?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockStore.java
```diff
@@ -121,11 +133,46 @@ public OptionalLong findNextMissingByte(long pos) throws IOException {
...
+          && !blockMetricsAndCacheHandler.isPresentInIndexCache(blockKey)) {
+        // The block is not in the index cache, so remove it from the block store
+        int range = blockKey.getRange().getLength();
```
fuatbasik, line 161: nit: length or size are probably better names here.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/util/BlockMetricsAndCacheHandler.java
```diff
@@ -16,29 +16,36 @@
...
 import software.amazon.s3.analyticsaccelerator.common.Metrics;
+import software.amazon.s3.analyticsaccelerator.io.physical.data.BlobStoreIndexCache;
 
 /**
  * Handles metrics updates for blocks in the Analytics Accelerator. This class provides thread-safe
  * operations for updating both blob-specific and aggregated metrics. It relies on the underlying
  * thread safety of the {@link Metrics} class for concurrent operations.
  */
-public class BlockMetricsHandler {
+public class BlockMetricsAndCacheHandler {
```
ahmarsuhail, line 26: In my opinion, I don't think you should change this class at all. the new methods here aren't doing anything that your indexCache isn't doing already. So just pass down your indexCache to your Block.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockStore.java
```diff
@@ -121,11 +133,46 @@ public OptionalLong findNextMissingByte(long pos) throws IOException {
...
+        try {
+          iterator.remove(); // Remove from the iterator as well
+          blockMetricsAndCacheHandler.reduceMetrics(MetricKey.MEMORY_USAGE, range);
+          LOG.debug(
+              "Removed block with key {}-{}-{} from block store during cleanup",
+              blockKey.getObjectKey().getS3URI(),
+              blockKey.getRange().getStart(),
+              blockKey.getRange().getEnd());
+
+        } catch (Exception e) {
```
fuatbasik, line 171: should we try to catch a better scoped exception here?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/util/BlockMetricsAndCacheHandler.java
```diff
@@ -56,4 +63,49 @@ public void updateMetrics(MetricKey key, long value) {
...
+  /**
+   * Reduces both blob-specific and aggregated metrics with the provided value. For memory usage
+   * metrics, updates both blob-specific and aggregated metrics. For other metrics, updates only the
+   * aggregated metrics. This method is thread-safe as it relies on the thread safety of the
+   * underlying {@link Metrics} implementation.
+   *
+   * @param key the metric key to be updated
+   * @param value the value to add to the metric
+   */
+  public void reduceMetrics(MetricKey key, long value) {
```
fuatbasik, line 76: wouldn't updateMetrics with negative value would do that?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/util/BlockMetricsAndCacheHandler.java
```diff
@@ -56,4 +63,49 @@ public void updateMetrics(MetricKey key, long value) {
...
+    if (key.equals(MetricKey.MEMORY_USAGE)) {
+      blobMetrics.reduce(key, value);
```
fuatbasik, line 78: update with negative value?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockStore.java
```diff
@@ -121,11 +133,46 @@ public OptionalLong findNextMissingByte(long pos) throws IOException {
...
+  public void add(BlockKey blockKey, Block block) {
     Preconditions.checkNotNull(block, "`block` must not be null");
 
-    this.blocks.add(block);
+    this.blocks.put(blockKey, block);
+  }
+
+  /**
+   * Cleans data from memory by removing blocks that are no longer needed. This method iterates
+   * through all blocks in memory and removes those that: 1. Have their data loaded AND 2. Are not
```
ahmarsuhail, line 146: why do they need to have their data loaded before they can be removed?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStoreIndexCache.java
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  /**
+   * Retrieves the range value associated with the specified block key from the cache, if it exists.
+   *
+   * @param blockKey the key whose associated range is to be returned
+   */
+  public void getIfPresent(BlockKey blockKey) {
```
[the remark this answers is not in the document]
ahmarsuhail: think we just used to record an access so rename to `recordBlockAccess()` or similar. update docs to state why you need to do this.

ahmarsuhail (changes requested):

input-stream/src/referenceTest/java/software/amazon/s3/analyticsaccelerator/reference/S3MockVsInMemoryReferenceTest.java
```diff
@@ -72,14 +75,27 @@ public class S3MockVsInMemoryReferenceTest extends StreamArbitraries {
   static void setup() {
     s3Client = createS3ClientV2(S3_MOCK.getHttpsEndpoint());
     // Initialise streams
+    Map<String, String> configMap = new HashMap<>();
+    long maxHeapBytes = Runtime.getRuntime().maxMemory();
+    double percentage = 0.01;
+    long capacityBytes = (long) (maxHeapBytes * percentage);
+
+    configMap.put("physicalio.memory.capacitybytes", String.valueOf(capacityBytes));
```
fuatbasik, line 83: Open Q: I know i recommended adding bytes there, now looking at it again, should we drop `bytes` or milliseconds` from the config flags?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStore.java
```diff
@@ -64,28 +71,51 @@ public BlobStore(
...
+  public void schedulePeriodicCleanup() {
+    maintenanceExecutor.scheduleAtFixedRate(
+        this::scheduleCleanupIfNotRunning,
+        configuration.getMemoryCleanupFrequencyMilliseconds(),
+        configuration.getMemoryCleanupFrequencyMilliseconds(),
+        TimeUnit.MILLISECONDS);
+  }
+
+  void scheduleCleanupIfNotRunning() {
+    if (metrics.get(MetricKey.MEMORY_USAGE) > configuration.getMemoryCapacityBytes()
```
ahmarsuhail, line 97: is there a reason why the scheduled cleanup only happen if memory usage > max_memory? why not just run it on schedule regardless?

fuatbasik (commented):
Thanks a lot @rajdchak. Awesome work. I am happy to approve once we have relevant Readme and Configuration documentation updated.

For future, I think we should think how can we better utilise the metrics to have more assertions.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
```diff
@@ -30,7 +31,8 @@
 @Builder
 @EqualsAndHashCode
 public class PhysicalIOConfiguration {
-  private static final int DEFAULT_CAPACITY_BLOB_STORE = 50;
+  private static final long DEFAULT_MEMORY_CAPACITY_BYTES = 2 * ONE_GB;
```
ahmarsuhail, line 34: do we think 2 * ONE_GB is enough as the default value? I know it's enough for our benchmarks, but is this really the default we want?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockStore.java
rajdchak: here we sync the blob map with index cache, if you follow the logic you will see we add the items in the index cache in the data loading step. If we do not add this check here we will end up prematurely removing items which were never added to index cache.

input-stream/src/referenceTest/java/software/amazon/s3/analyticsaccelerator/reference/S3MockVsInMemoryReferenceTest.java
ahmarsuhail: +1, drop both. should just be `max.memory` or similar. docs to say what unit these values they use.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockStore.java
rajdchak: here we sync the blob map with index cache, if you follow the logic you will see we add the items in the index cache in the data loading step. If we do not add this check here we will end up prematurely removing items which were never added to index cache.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/util/BlockMetricsAndCacheHandler.java
rajdchak: yeah they would, but in last PR @ahmarsuhail asked not to pass around negative values in all methods,  instead have the reduce method in the metrics class which will actually negate the value.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/util/BlockMetricsAndCacheHandler.java
rajdchak: same comment as above. If you follow the code flow ultimately the reduce method will negate this value.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/parquet/ParquetReadTailTask.java
```diff
@@ -75,11 +75,11 @@ public FileTail readFileTail() {
               ParquetUtils.getFileTailRange(logicalIOConfiguration, 0, contentLength);
           if (tailRangeOptional.isPresent()) {
             Range tailRange = tailRangeOptional.get();
-            int tailLength = (int) tailRange.getLength();
+            int tailLength = tailRange.getLength();
             try {
               byte[] fileTail = new byte[tailLength];
               physicalIO.readTail(fileTail, 0, tailLength);
-              return new FileTail(ByteBuffer.wrap(fileTail), (int) tailRange.getLength());
+              return new FileTail(ByteBuffer.wrap(fileTail), tailRange.getLength());
```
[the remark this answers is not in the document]
stubz151: Think this is still valid.

input-stream/src/referenceTest/java/software/amazon/s3/analyticsaccelerator/reference/S3MockVsInMemoryReferenceTest.java
rajdchak: why to drop? I think it provides a better understanding of the unit.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStore.java
rajdchak: No reason as such I just thought may be not waste cpu cycles doing cleanup. But I can remove this condition if we want. Shouldn't impact much.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
rajdchak: I am not sure right now on how to fix on a default value other than following the results from our benchmarks.  Customers can increase if they want is what I feel. @fuatbasik what do you feel?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStore.java
```diff
@@ -64,28 +71,51 @@ public BlobStore(
...
-                      "Current memory usage of blobMap in bytes after eviction is: {}",
-                      metrics.get(MetricKey.MEMORY_USAGE));
-                }
-                return shouldRemove;
-              }
+    this.blobMap = Collections.synchronizedMap(new LinkedHashMap<ObjectKey, Blob>());
+    this.indexCache = new BlobStoreIndexCache(configuration);
+    this.maintenanceExecutor =
+        Executors.newSingleThreadScheduledExecutor(
+            r -> {
```
stubz151, line 78: can we give r a better name here, same as t below.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManager.java
```diff
@@ -17,9 +17,7 @@
 
 import java.io.Closeable;
 import java.io.IOException;
-import java.util.List;
-import java.util.Optional;
-import java.util.OptionalLong;
+import java.util.*;
```
stubz151, line 20: Check * imports in this class

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockStore.java
stubz151: Why do we want to use * imports here?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/util/BlockMetricsAndCacheHandler.java
rajdchak: I prefer to keep this class as a separate wrapper/handler for the Block, metrics, and index cache, rather than extending the Block constructor. This approach provides better separation of concerns and encapsulation, making the code more maintainable and cleaner.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
rajdchak: yes this is to update the access to the key, yes I can move this to a different method.

input-stream/src/referenceTest/java/software/amazon/s3/analyticsaccelerator/property/InMemoryS3SeekableInputStream.java
```diff
@@ -43,8 +48,19 @@ public InMemoryS3SeekableInputStream(String bucket, String key, int len) throws
     S3URI s3URI = S3URI.of(bucket, key);
     ObjectClient objectClient = new InMemoryObjectClient(len);
 
-    S3SeekableInputStreamFactory factory =
-        new S3SeekableInputStreamFactory(objectClient, S3SeekableInputStreamConfiguration.DEFAULT);
+    Map<String, String> configMap = new HashMap<>();
+    long maxHeapBytes = Runtime.getRuntime().maxMemory();
+    double percentage = 0.01;
+    long capacityBytes = (long) (maxHeapBytes * percentage);
```
stubz151, line 54: think we can move this logic to a separate method like getDefaultMemoryConfigs or something like that and maybe move it to the physical io config class.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
```diff
@@ -191,9 +175,15 @@ private void generateSourceAndData() throws IOException {
             this.source.thenApply(
                 objectContent -> {
                   try {
-                    this.metricsHandler.updateMetrics(MetricKey.MEMORY_USAGE, range.getLength());
+                    int blockRange = blockKey.getRange().getLength();
+                    this.blockMetricsAndCacheHandler.updateMetrics(
+                        MetricKey.MEMORY_USAGE, blockRange);
+                    this.blockMetricsAndCacheHandler.putInIndexCache(blockKey, blockRange);
```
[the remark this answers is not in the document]
rajdchak: I don't think there is any problem in doing it after as well. This doesn't matter much actually doing before or after. The main thing to be avoided was doing all this just after creating the block obj and before joining on the get request.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
```diff
@@ -163,7 +198,7 @@ private PhysicalIOConfiguration(
       double sequentialPrefetchSpeed,
       long blockReadTimeout,
       int blockReadRetryCount) {
-    Preconditions.checkArgument(blobStoreCapacity > 0, "`blobStoreCapacity` must be positive");
+    Preconditions.checkArgument(memoryCapacityBytes > 0, "`blobStoreCapacity` must be positive");
```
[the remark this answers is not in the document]
rajdchak: I think I missed adding the checks, will do that.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
```diff
@@ -41,11 +43,33 @@ public class PhysicalIOConfiguration {
...
+  private static final String MEMORY_CAPACITY_BYTES_KEY = "memory.capacitybytes";
+
+  /**
+   * Capacity, in blobs. {@link
+   * PhysicalIOConfiguration#DEFAULT_MEMORY_CLEANUP_FREQUENCY_MILLISECONDS} by default.
+   */
+  @Builder.Default
+  private int memoryCleanupFrequencyMilliseconds = DEFAULT_MEMORY_CLEANUP_FREQUENCY_MILLISECONDS;
+
+  private static final String MEMORY_CLEANUP_FREQUENCY_MILLISECONDS_KEY =
```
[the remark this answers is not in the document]
rajdchak: why not have the unit? isn't it for someone to understand rather than referring to document everytime?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStore.java
```diff
@@ -124,15 +160,22 @@ public boolean evictKey(ObjectKey objectKey) {
...
+        maintenanceExecutor.shutdownNow();
+      }
+      blobMap.forEach((k, v) -> v.close());
+      indexCache.cleanUp();
+      long hits = metrics.get(MetricKey.CACHE_HIT);
+      long miss = metrics.get(MetricKey.CACHE_MISS);
+      LOG.debug(
+          "Cache Hits: {}, Misses: {}, Hit Rate: {}%",
+          hits, miss, MetricComputationUtils.computeCacheHitRate(hits, miss));
+    } catch (Exception e) {
```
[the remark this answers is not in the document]
rajdchak: what is the concern with not doing blobMap=null ?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Blob.java
```diff
@@ -71,18 +73,27 @@ public Blob(
...
+      return blockManager.getBlock(pos).get().read(pos);
+    } finally {
+      lock.readLock().unlock();
+    }
+  }
 
-    return blockManager.getMemoryUsageOfBlob();
+  /** clean up blob */
+  public final void asyncCleanup() {
```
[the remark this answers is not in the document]
rajdchak: ack

stubz151 (changes requested):
Just had a few small resolved comments that I'm not sure why they were resolved.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
rajdchak: ack

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/parquet/ParquetReadTailTask.java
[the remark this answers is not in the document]
rajdchak: I didn't make this change though, I am not sure why was it done!

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockStore.java
rajdchak: yeah I  will update this

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
fuatbasik: sorry if it created confusion, what is the reason for this if/else block here?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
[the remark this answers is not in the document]
ahmarsuhail: usually config values like this don't have values. look at S3A as an example, `fs.s3a.connection.part.upload.timeout` from https://github.com/apache/hadoop/blob/trunk/hadoop-tools/hadoop-aws/src/site/markdown/tools/hadoop-aws/connecting.md

you don't need the units, anyone setting this will refer to the docs. Ideally the convention is you allow the unit in your value, so you can pass down values like: 400MB, 2G, 1T. so if you have the value in your config, you won't be able to

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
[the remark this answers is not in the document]
ahmarsuhail: cool let's move it to after. i want to make some changes here which mean it might start making a difference where these are.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/util/BlockMetricsAndCacheHandler.java
ahmarsuhail: I actually think the opposite :) to me having two separate classes is a better separation. happy to discuss more

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
rajdchak: this is to update the access to the key in the index cache as well. I can move this logic to a different method.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
```diff
@@ -41,11 +43,33 @@ public class PhysicalIOConfiguration {
...
   private static final long DEFAULT_BLOCK_READ_TIMEOUT = 30_000;
   private static final int DEFAULT_BLOCK_READ_RETRY_COUNT = 20;
+  private static final int DEFAULT_MEMORY_CLEANUP_FREQUENCY_MILLISECONDS = 5000;
 
-  /** Capacity, in blobs. {@link PhysicalIOConfiguration#DEFAULT_CAPACITY_BLOB_STORE} by default. */
-  @Builder.Default private int blobStoreCapacity = DEFAULT_CAPACITY_BLOB_STORE;
+  /**
+   * Capacity, in blobs. {@link PhysicalIOConfiguration#DEFAULT_MEMORY_CAPACITY_BYTES} by default.
+   */
+  @Builder.Default private long memoryCapacityBytes = DEFAULT_MEMORY_CAPACITY_BYTES;
```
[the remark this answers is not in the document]
rajdchak: okay but I think adding the units makes it easier to understand?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockStore.java
ahmarsuhail: what does this mean for blocks that are never accessed? eg: Block was created (we prefetched inaccurately and made extra get requests), but it was never read

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStore.java
ahmarsuhail: what do you think is the right thing to do here? i think if we should remove this clause, because it means we won't do the cleanup proactively. Eg: if someone sets max memory to 100GB, we won't clear the cache and likely hit OOMs

rajdchak: cool, I can remove this check

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockStore.java
rajdchak: we don't the data also right in those cases, then it doesn't occupy much memory right?

rajdchak: here the purpose is not prevent get requests but is to limit the memory.

rajdchak: if we have to address the inaccurate prefetch cases, that should be done as part of a different work then.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/util/BlockMetricsAndCacheHandler.java
rajdchak: If you see, index and metrics classes are actually separate, this class is just a wrapper around them so that blocks can access the methods in those two classes via this handler. In future we can extend this handler to perform various other functionalities for the blocks.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
rajdchak: As discusses, we are sticking to this value. We will create a new issue to decide on a default value per instance type.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStore.java
rajdchak: ack

stubz151 (approved):

README.md
```diff
@@ -134,6 +134,21 @@ These optimizations are:
 * Predictive column prefetching - The library tracks recent columns being read using parquet metadata. When
   subsequent Parquet files which have these columns are opened, the library will prefetch these columns. For example, if columns `x` and `y` are read from `A.parquet` , and then `B.parquet` is opened, and it also contains columns named `x` and `y`, the library will prefetch them asynchronously.
 
+## Memory Used by Library
+
+* The Analytics Accelerator Library uses a memory structure called blobstore to store data fetched from S3.
```
fuatbasik, line 139: Analytics Accelerator Library for `Amazon S3`

README.md
```diff
@@ -134,6 +134,21 @@ These optimizations are:
...
+* To keep the memory usage under a limit, the library implements a Caffeine index cache to store indexes of all blocks across all blobs. Each block is indexed using a BlockKey (combination of ObjectKey [S3Uri + Etag] and Range [start and end byte]).
```
fuatbasik, line 140: i do not think you should mention caffeine here at all.

fuatbasik: i found this part a bit deep into implementation. How about something like this:

AAL fetches data from S3 in blocks of bytes and keeps them in memory. These memory items are expired after a configurable TTL (X seconds by default). Also, total memory usage is limited to a configurable threshold which is set to 2GB by default. When memory usage exceeds this threshold blocks to be removed are identifed using XXX algorithm, implemented by Caffeine library (link here paper & caffeine lib). Yet, removal is done using an async. process that is running every X seconds, meaning that memory used by AAL might go beyond the set threshold for X seconds. Setting a higher frequency will reduce time AAL might use more than allowed memory. 

You can change TTL, memory usage threshold and cleanup frequency as follows:

- Flag 1
- Flag 2
- Flag 3

fuatbasik: A follow-up question i have, if customer sets memory.cleanup.frequency  to 0, will blocks identified for eviction will be evicted immediately and AAL will always remain using less than max.memory.limit bytes?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Blob.java
```diff
@@ -42,6 +43,24 @@ public class Blob implements Closeable {
   private final ObjectMetadata metadata;
   private final Telemetry telemetry;
 
+  /**
+   * The ReentrantReadWriteLock manages concurrent access between read operations and eviction:<br>
+   *
+   * <p>Read Lock (used in read() and read(byte[], int, int, long) methods):<br>
```
fuatbasik, line 49: do you need <p> and <br> tags? I havent seen them being used anywhere else in our comments.

fuatbasik (commented):
Thanks @rajdchak . Sorry not called this out in the first review but shall we add more tests around corner cases. For example what happens if customers set max.memory.limit to 0 or cleanup frequency to 0?

rajdchak:
> Thanks @rajdchak . Sorry not called this out in the first review but shall we add more tests around corner cases. For example what happens if customers set max.memory.limit to 0 or cleanup frequency to 0?

We have precondition checks which checks whether these two values are greater than 0 or not.

README.md
rajdchak: We have precondition checks which checks whether these values are greater than 0 or not.

rajdchak: Also we can't set frequency to something less than equals to 0 as the ScheduledThreadPoolExecutor will throw IllegalArgumentException.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Blob.java
rajdchak: we use tags in many places in our code base. Without the tags the comments don't appear in a nice format.

README.md
```diff
@@ -135,20 +135,15 @@ These optimizations are:
...
-* Cleanup frequency can be set using the key `memory.cleanup.frequency` by default which is `5s`. Time period (in milliseconds) between consecutive BlobStore eviction cycles.
+Analytics Accelerator Library for Amazon S3 implements a best-effort memory limiting mechanism. The library fetches data from S3 in blocks of bytes and keeps them in memory. Memory management is achieved through a dual strategy combining Time-to-Live (TTL) and maximum memory threshold.
+When time to live or memory usage exceeds the configured threshold, blocks to be removed are identified using [Time_based_eviction](https://github.com/ben-manes/caffeine/wiki/Eviction#time-based) and [Window TinyLfu algorithm](https://github.com/ben-manes/caffeine/wiki/Efficiency) respectively, implemented by [Caffeine library](https://github.com/ben-manes/caffeine/blob/master/README.md). Removal is done using an async process that runs every `X` seconds, meaning memory usage might temporarily exceed the threshold. This overflow period can be minimized by increasing the cleanup frequency, though at the cost of higher CPU utilization.
```
fuatbasik, line 139: what about instead of Removal is done using an async process that runs every `X` seconds

`Removal is done using an async process with a configurable frequency (which is set to 5 seconds by default). This means memory usage might ...`

fuatbasik (approved):
@rajdchak thanks a lot. Approved!

ahmarsuhail (approved):

README.md
rajdchak: As discussed we will think about the 0 scenario in a follow up task.

PR #257, opened 2025-04-17
Added stats to record memory usage and cache hit/miss

rajdchak:
## Description of change
Added stats to record memory usage and cache hit/miss

---


common/src/main/java/software/amazon/s3/analyticsaccelerator/stats/MemoryUsageStats.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Tracks memory usage statistics across all blob maps in the system. This class provides
+ * thread-safe methods to record, retrieve, and reset memory usage statistics.
+ */
+public class MemoryUsageStats {
+  /** Atomic counter for tracking total memory usage across all blob maps. */
+  private static final AtomicLong memoryUsageAcrossBlobMap = new AtomicLong(0);
```
fuatbasik, line 26: nit: Maybe drop BlobMap from the name and call it memoryUsage only. This is part of Common package which doesn't know about the BlobMap at all.

common/src/test/java/software/amazon/s3/analyticsaccelerator/request/ObjectMetadataTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+@SuppressFBWarnings(
+    value = "NP_NONNULL_PARAM_VIOLATION",
+    justification = "Null values are intentionally tested")
+class ObjectMetadataTest {
```
fuatbasik, line 28: wait a sec. We didn't have this before? Thanks a ton!

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStream.java
```diff
@@ -46,6 +48,7 @@ public class S3SeekableInputStream extends SeekableInputStream {
 
   private static final String OPERATION_STREAM_CLOSE = "seekablestream.close";
   private final long streamBirth = System.nanoTime();
+  private static final Logger LOG = LoggerFactory.getLogger(S3SeekableInputStream.class);
```
fuatbasik, line 51: Are changes to this class leftover from other testing?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStore.java
```diff
@@ -57,7 +63,23 @@ public BlobStore(
...
               @Override
               protected boolean removeEldestEntry(final Map.Entry<ObjectKey, Blob> eldest) {
-                return this.size() > configuration.getBlobStoreCapacity();
+                boolean shouldRemove = this.size() > configuration.getBlobStoreCapacity();
+                if (shouldRemove) {
+                  Blob blobToRemove = eldest.getValue();
+                  long blobMemoryToSubtract = 0;
+
+                  // Calculate memory used by this blob
+                  List<Block> blocks = blobToRemove.getBlockManager().getBlockStore().getBlocks();
```
fuatbasik, line 72: wait a sec. do we need to move block by block here when removing? Can we add a filed to Blob (sumOfBytes) and update it when adding, and just substract here when removing?

fuatbasik: i think this is even risky to iterate through blocks out of control of the BlockManager/BlockStore, worst case we should write a method to blockStore `getTotalSize` and call it from here.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockStore.java
```diff
@@ -34,7 +36,7 @@ public class BlockStore implements Closeable {
 
   private final ObjectKey s3URI;
   private final ObjectMetadata metadata;
-  private final List<Block> blocks;
+  @Getter private final List<Block> blocks;
```
fuatbasik, line 39: No! there is a reason blocks are private and there is no getters. We should not have direct-access to store without the manager.

fuatbasik (changes requested):
Thanks a lot @rajdchak this is awesome. I think we should not violate BlockManager-BlockStore encapsulation and avoid accessing to Store and Blocks directly. Other than these an awesome change.

common/src/main/java/software/amazon/s3/analyticsaccelerator/stats/MemoryUsageStats.java
rajdchak: I know but this field tracks only memory usage of blobMap, if I change the name to memoryUsage it will be misleading and look like this value is the total memory usage of our library

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Blob.java
```diff
@@ -72,7 +73,9 @@ public Blob(
   public int read(long pos) throws IOException {
     Preconditions.checkArgument(pos >= 0, "`pos` must be non-negative");
     blockManager.makePositionAvailable(pos, ReadMode.SYNC);
-    return blockManager.getBlock(pos).get().read(pos);
+    Block block = blockManager.getBlock(pos).get();
```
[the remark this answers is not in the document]
rajdchak: This is unrelated change will revert

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStream.java
rajdchak: ack

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStore.java
rajdchak: changed to compute this at blob level

rajdchak:
> https://github.com/ben-manes/caffeine/wiki/Statistics question why aren't we using this?

Right now the blob map is not a caffeine cache. When we have the memory manager changes we are adopting caffeine but that is only to store the indexes, we will continue to have the blopmap and other structures as we have today.

stubz151:
> > https://github.com/ben-manes/caffeine/wiki/Statistics question why aren't we using this?
> 
> Right now the blob map is not a caffeine cache. When we have the memory manager changes we are adopting caffeine but that is only to store the indexes, we will continue to have the blopmap and other structures as we have today. Instead of trusting the caffeine stats these stats are more reliable.

Ahh makes perfect sense. Sorry for the miss.

common/src/main/java/software/amazon/s3/analyticsaccelerator/stats/MemoryUsageStats.java
fuatbasik: I am sorry but that doesn't make much sense to me. Why not create an instance of this object (make fields non-static) in the blobMap and call it blobMapMemoryTracker and have increment decrement methods. I do not agree that the fieldName of this class should be consumer specific.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStore.java
```diff
@@ -34,7 +38,8 @@
     justification =
         "Inner class is created very infrequently, and fluency justifies the extra pointer")
 public class BlobStore implements Closeable {
-  private final Map<ObjectKey, Blob> blobMap;
+  @Getter private final Map<ObjectKey, Blob> blobMap;
```
fuatbasik, line 41: i do not think you should have public access to blobmap. Any operation should be done through BlobStore/BlobManager.

fuatbasik (changes requested):
Hi @rajdchak ,
I had another look and put some more comments. In my opinion, you should try to avoid statics and global variables as much as possible as they make it harder to debug and reason about the correctness. Another thing is, In the current structure we have States (BlobStore/BlockStore) and StateManagers(BlobManager, BlockManager). We should avoid accessing state out of control of the Manager. So if you need to access BlobStore, it should be through BlobManager.


My main concerns are
1/ You do not need any Statics at CacheStats and MemoryStats. 
2/ You do not need to pass a global CacheStats or MemoryStats. You can aggregate all metrics at a higher level therefore, it is OK to create 1 CacheStats and 1 MemoryStats metric per Blob and have AggregatingCacheStats and AggregatingMemoryStats at BlobStore level. When metrics asked, BlobStore will aggregate metrics per Blob. Current version creates an interesting dependency in between the classes. 

I also like to hear from others. @ahmarsuhail @stubz151

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockStore.java
```diff
@@ -61,7 +62,13 @@ public BlockStore(ObjectKey objectKey, ObjectMetadata metadata) {
   public Optional<Block> getBlock(long pos) {
     Preconditions.checkArgument(0 <= pos, "`pos` must not be negative");
 
-    return blocks.stream().filter(b -> b.contains(pos)).findFirst();
+    Optional<Block> block = blocks.stream().filter(b -> b.contains(pos)).findFirst();
+    if (block.isPresent()) {
+      CacheStats.recordHit();
```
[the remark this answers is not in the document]
rajdchak: if block is there, but block data is still being fetched, we cannot record that as a miss right. So, imo we should recording a hit if a block is present should be fine.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStore.java
rajdchak: I just needed this for the tests and not any other code logic, l will change this.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStreamFactory.java
```diff
@@ -177,8 +179,14 @@ void storeObjectMetadata(S3URI s3URI, ObjectMetadata metadata) {
...
   @Override
   public void close() throws IOException {
+    LOG.debug(
+        "Cache Hits: {}, Misses: {}, Hit Rate: {}%",
+        CacheStats.getHits(), CacheStats.getMisses(), CacheStats.getHitRate() * 100);
+
     this.objectMetadataStore.close();
     this.objectBlobStore.close();
     this.telemetry.close();
+    MemoryUsageStats.resetStats();
```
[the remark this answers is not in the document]
rajdchak: I had just made the things static so that I do not have to pass around the objects in all the classes and it was faster to make the changes during testing. I will make this non static.

rajdchak:
> Hi @rajdchak , I had another look and put some more comments. In my opinion, you should try to avoid statics and global variables as much as possible as they make it harder to debug and reason about the correctness. Another thing is, In the current structure we have States (BlobStore/BlockStore) and StateManagers(BlobManager, BlockManager). We should avoid accessing state out of control of the Manager. So if you need to access BlobStore, it should be through BlobManager.
> 
> My main concerns are 1/ You do not need any Statics at CacheStats and MemoryStats. 2/ You do not need to pass a global CacheStats or MemoryStats. You can aggregate all metrics at a higher level therefore, it is OK to create 1 CacheStats and 1 MemoryStats metric per Blob and have AggregatingCacheStats and AggregatingMemoryStats at BlobStore level. When metrics asked, BlobStore will aggregate metrics per Blob. Current version creates an interesting dependency in between the classes.
> 
> I also like to hear from others. @ahmarsuhail @stubz151

I will make things non static.

However, if we do not maintain the exact memory usage numbers at the global level, then everytime during eviction we will need to loop over blobs and aggregate to measure, do we want to do that? (edited) 
When the memory manager changes will be in, we will run evictions at regular intervals and it will lead to computing aggregations frequently. Also with the memory manager changes, we do not need to maintain the memory usage at a blob level because we will not evict entire blob in one go. Right now we need the memory usage at blob level just to subtract the size of the blob during eviction.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockStore.java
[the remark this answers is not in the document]
fuatbasik: Going forward, we probably need to think about this in more detail. Having block ready but bytes not fetched are probably is not good enough.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStore.java
```diff
@@ -105,6 +136,33 @@ public int blobCount() {
...
+    double hitRate = hitsAndMiss == 0 ? 0 : (double) hits / hitsAndMiss;
+    LOG.debug(
+        "Cache Hits: {}, Misses: {}, Hit Rate: {}%",
+        cacheHits.get(), cacheMiss.get(), hitRate * 100);
     blobMap.forEach((k, v) -> v.close());
   }
+
+  /** Callbacks from block manager instances. */
+  public class BlobStoreCallbacks {
+
```
rajdchak, line 150: With memory manager changes this class will be extended to have methods to update the index cache.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStore.java
```diff
@@ -52,15 +62,30 @@ public BlobStore(
       @NonNull PhysicalIOConfiguration configuration) {
     this.objectClient = objectClient;
     this.telemetry = telemetry;
+    this.blobStoreMemoryUsage = new AtomicLong(0);
+    this.cacheHits = new AtomicInteger(0);
```
ahmarsuhail, line 66: I read don't think these values belong here. Taking a long term view, you want a IoStats interface + class, which gets updated. Look at how S3A does it, as an example. But here's how it should work. 

You have a IoStats class, you instantiate it when you create a factory. This is the class that holds all values. Currently you're only measuring cache hits. in the future we will want to know how many GETs, how many bytes we read etc etc. if we start putting these values in our main classes this is going to get very messy. 

All classes that want to update statistics, should use callbacks to update these IoStats. 

I understand that here you want statistics per blobMap, S3A has a way of doing this too. it has this concept called IoStreamStatistics, those are statistics per IoStream. you probably want something similar here. And then on the blobstore close, you can push those up to a factory statistics. 

I know this more work than you signed up for, and it's not as straightforward. But I don't think we should do it this way.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStore.java
```diff
@@ -105,6 +136,35 @@ public int blobCount() {
...
+    long hitsAndMiss = hits + cacheMiss.get();
+    double hitRate = hitsAndMiss == 0 ? 0 : (double) hits / hitsAndMiss;
+    LOG.debug(
+        "Cache Hits: {}, Misses: {}, Hit Rate: {}%",
+        cacheHits.get(), cacheMiss.get(), hitRate * 100);
     blobMap.forEach((k, v) -> v.close());
   }
+
+  /** Callbacks from block manager instances. */
+  public class BlobStoreCallbacks {
```
ahmarsuhail, line 149: similar to above, we should put all IoStats methods in one place, and they shouldn't be in this class.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManager.java
```diff
@@ -230,4 +243,28 @@ private long truncatePos(long pos) {
   public void close() {
     blockStore.close();
   }
+
+  /** Callbacks from block instances. */
+  public class BlockManagerCallback {
```
ahmarsuhail, line 248: having that IoStats class should you don't need to define callbacks twice

ahmarsuhail (changes requested):

common/src/main/java/software/amazon/s3/analyticsaccelerator/common/Metrics.java
```diff
@@ -0,0 +1,67 @@
...
+  private final Map<MetricKey, AtomicLong> metrics = new ConcurrentHashMap<>();
+
+  /**
+   * Adds the specified delta to the metric identified by the given key. If the metric doesn't
+   * exist, it will be created with an initial value of 0 before adding the delta.
+   *
+   * @param key the metric key to identify which metric to update
+   * @param delta the value to add to the metric (can be negative for decrements)
+   */
+  public void add(MetricKey key, long delta) {
```
fuatbasik, line 42: nit: A helper toString() might help when you are logging the metrics. Not for now, we can add later.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
```diff
@@ -186,8 +195,11 @@ private void generateSourceAndData() throws IOException {
             this.source.thenApply(
                 objectContent -> {
                   try {
-                    return StreamUtils.toByteArray(
-                        objectContent, this.objectKey, this.range, this.readTimeout);
+                    byte[] bytes =
+                        StreamUtils.toByteArray(
+                            objectContent, this.objectKey, this.range, this.readTimeout);
+                    metricsCallback.accept(MetricKey.MEMORY_USAGE, range.getLength());
```
fuatbasik, line 201: nit: OK for now, in the future, might want to take values from here and do more businessLogic on the BlockManager side.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManager.java
```diff
@@ -48,6 +52,8 @@ public class BlockManager implements Closeable {
   private final PhysicalIOConfiguration configuration;
   private final RangeOptimiser rangeOptimiser;
   private StreamContext streamContext;
+  @Getter private AtomicLong blobMemoryUsage;
```
fuatbasik, line 55: you can even avoid this right, all you need is another metrics instance specific to this blob.

fuatbasik: why do you need getter for this field?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManager.java
```diff
@@ -85,18 +94,21 @@ public BlockManager(
...
       @NonNull PhysicalIOConfiguration configuration,
+      @NonNull Metrics metrics,
       StreamContext streamContext) {
     this.objectKey = objectKey;
     this.objectClient = objectClient;
     this.metadata = metadata;
     this.telemetry = telemetry;
     this.configuration = configuration;
-    this.blockStore = new BlockStore(objectKey, metadata);
+    this.metrics = metrics;
```
fuatbasik, line 104: nit: maybe call it AggregatingMetrics and create another one blobMetrics and update both on your callback.

fuatbasik (dismissed):
thanks a lot @rajdchak . Approved!

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStore.java
```diff
@@ -105,6 +128,11 @@ public int blobCount() {
   /** Closes the {@link BlobStore} and frees up all resources it holds. */
   @Override
   public void close() {
+    long hits = metrics.get(MetricKey.CACHE_HIT);
```
ahmarsuhail, line 131: nit:  can we move this calculation into a utils class?

stubz151 (approved):

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStore.java
```diff
@@ -35,29 +40,47 @@
...
               @Override
               protected boolean removeEldestEntry(final Map.Entry<ObjectKey, Blob> eldest) {
-                return this.size() > configuration.getBlobStoreCapacity();
+                boolean shouldRemove = this.size() > configuration.getBlobStoreCapacity();
+                if (shouldRemove) {
+                  LOG.debug(
+                      "Current memory usage of blobMap in bytes before eviction is: {}",
+                      metrics.get(MetricKey.MEMORY_USAGE));
+                  Blob blobToRemove = eldest.getValue();
+                  metrics.add(MetricKey.MEMORY_USAGE, -blobToRemove.getMemoryUsageOfBlob());
```
ahmarsuhail, line 78: instead of the `-` sign, can we add a `metrics.reduce` method or similar. that `-` makes me a bit nervous..too easy to remove and will change the behaviour

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStore.java
ahmarsuhail: actually, maybe you can just have a average() method in your metrics class, that takes the two statistics you want to calc the average for. really don't want this logic in this class

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManager.java
ahmarsuhail: +1, don't think you need the getter. 

Would also prefer the metrics class rather than the atomic long.. but in two minds. can't figure out if that's an overkill here

ahmarsuhail (changes requested):
just a couple of minor changes!

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManager.java
rajdchak: yeah I didn't pay much attention to this because I will be removing this memory usage from here when i have the memory manager changes because we do not need to keep a track of blob level memory usage then.

rajdchak: Needed a getter for this call from blob
```

  /**
   * Returns the memory used by the blob
   *
   * @return the memory used by the blob
   */
  public long getMemoryUsageOfBlob() {
    return blockManager.getBlobMemoryUsage().get();
  }
```

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlobStore.java
rajdchak: this is actually not an average calculation, may be we can call it ratio?

rajdchak: moving this to utils class as moving it to Metrics class specifically to calculate hit rate doesn't make sense

fuatbasik (approved):

ahmarsuhail (approved):
+1, LGTM

fuatbasik (approved):

PR #258, opened 2025-04-22
Move small object prefetching to physicalIO

vaibhav5140:
## Description of change
Moves small object prefetching from logical to physical layer for format-agnostic optimization.

No. Small object prefetching functionality is moved internally from logical to physical layer while maintaining existing behaviour.

Yes. Added new configuration options in PhysicalIOConfiguration:
- smallObjectsPrefetchingEnabled
- smallObjectSizeThreshold These enable small object prefetching configuration at physical layer.
---

### DIFF
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -50,8 +50,6 @@ Options under `<CONNECTOR_PREFIX>.logicalio.`
 | `prefetch.file.page.index.size`       | `1MB`                 | Size of page index to prefetch for regular files                           |
 | `prefetch.large.file.page.index.size` | `8MB`                 | Size of page index to prefetch for large files                             |
 | `large.file.size`                     | `1GB`                 | Threshold to consider a file as large                                      |
-| `small.objects.prefetching.enabled`   | `true`                | Controls prefetching for small objects                                     |
-| `small.object.size.threshold`         | `3MB`                 | Size threshold for small object prefetching                                |
 | `parquet.metadata.store.size`         | `45`                  | Size of the parquet metadata store                                         |
 | `max.column.access.store.size`        | `15`                  | Maximum size of column access store                                        |
 | `parquet.format.selector.regex`       | `^.*.(parquet\|par)$` | Regex pattern to identify parquet files                                    |
@@ -60,19 +58,20 @@ Options under `<CONNECTOR_PREFIX>.logicalio.`
 ## Physical IO Configuration
 Options under `<CONNECTOR_PREFIX>.physicalio.`
 
-| Option                    | Default | Description                                                                        |
-|---------------------------|---------|------------------------------------------------------------------------------------|
-| `metadatastore.capacity`  | `50`    | Capacity of the metadata store                                                     |
-| `blocksizebytes`          | `8MB`   | Size of blocks for data transfer                                                   |
-| `readaheadbytes`          | `64KB`  | Number of bytes to read ahead                                                      |
-| `maxrangesizebytes`       | `8MB`   | Maximum size of range requests                                                     |
-| `partsizebytes`           | `8MB`   | Size of individual parts for transfer                                              |
-| `sequentialprefetch.base` | `2.0`   | Base factor for sequential prefetch sizing                                         |
-| `sequentialprefetch.speed`| `1.0`   | Speed factor for sequential prefetch growth                                        |
-| `max.memory.limit`        | `2GB`   | Maximum memory limit (in bytes) that BlobStore can utilize for storage             |
-| `cache.timeout`           | `1s`    | Time duration (in milliseconds) a block remains in BlobStore after its last access |
-| `memory.cleanup.frequency`| `5s`    | Time period (in milliseconds) between consecutive BlobStore eviction cycles        |
-
+| Option                              | Default | Description                                                                        |
+|-------------------------------------|---------|------------------------------------------------------------------------------------|
+| `metadatastore.capacity`            | `50`    | Capacity of the metadata store                                                     |
+| `blocksizebytes`                    | `8MB`   | Size of blocks for data transfer                                                   |
+| `readaheadbytes`                    | `64KB`  | Number of bytes to read ahead                                                      |
+| `maxrangesizebytes`                 | `8MB`   | Maximum size of range requests                                                     |
+| `partsizebytes`                     | `8MB`   | Size of individual parts for transfer                                              |
+| `sequentialprefetch.base`           | `2.0`   | Base factor for sequential prefetch sizing                                         |
+| `sequentialprefetch.speed`          | `1.0`   | Speed factor for sequential prefetch growth                                        |
+| `max.memory.limit`                  | `2GB`   | Maximum memory limit (in bytes) that BlobStore can utilize for storage             |
+| `cache.timeout`                     | `1s`    | Time duration (in milliseconds) a block remains in BlobStore after its last access |
+| `memory.cleanup.frequency`          | `5s`    | Time period (in milliseconds) between consecutive BlobStore eviction cycles        |
+| `small.objects.prefetching.enabled` | `true`  | Controls prefetching for small objects                                             |
+| `small.object.size.threshold`       | `8MB`   | Size threshold for small object prefetching                                        |
 ## Telemetry Configuration
 Options under `<CONNECTOR_PREFIX>.telemetry.`
 
--- a/input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/IntegrationTestBase.java
+++ b/input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/IntegrationTestBase.java
@@ -18,10 +18,12 @@ package software.amazon.s3.analyticsaccelerator.access;
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static software.amazon.s3.analyticsaccelerator.access.ChecksumAssertions.assertChecksums;
 import static software.amazon.s3.analyticsaccelerator.util.Constants.ONE_KB;
+import static software.amazon.s3.analyticsaccelerator.util.Constants.ONE_MB;
 
 import java.io.IOException;
 import java.io.InputStream;
@@ -319,6 +321,60 @@ public abstract class IntegrationTestBase extends ExecutionBase {
       // Shutdown. Wait for termination indefinitely - we expect it to always complete
       executorService.shutdown();
       assertTrue(executorService.awaitTermination(Long.MAX_VALUE, TimeUnit.SECONDS));
+    }
+  }
+
+  /**
+   * Tests the prefetching behavior for S3 objects with focus on cache verification. For objects
+   * smaller than 8MB, the entire object should be prefetched into cache on first read. For objects
+   * larger than 8MB, data should be fetched from S3 as needed.
+   *
+   * @param s3ClientKind the type of S3 client to use
+   * @param s3Object the S3 object to test with
+   * @param streamReadPatternKind the pattern to use for reading the stream
+   * @param AALInputStreamConfigurationKind the configuration for the input stream
+   * @throws IOException if an I/O error occurs during the test
+   */
+  // TODO: Update test to assert on GET request metrics once we have them
+  protected void testSmallObjectPrefetching(
+      @NonNull S3ClientKind s3ClientKind,
+      @NonNull S3Object s3Object,
+      @NonNull StreamReadPatternKind streamReadPatternKind,
+      @NonNull AALInputStreamConfigurationKind AALInputStreamConfigurationKind)
+      throws IOException {
+
+    try (S3AALClientStreamReader s3AALClientStreamReader =
+        this.createS3AALClientStreamReader(s3ClientKind, AALInputStreamConfigurationKind)) {
+
+      // First stream
+      S3SeekableInputStream stream = s3AALClientStreamReader.createReadStream(s3Object);
+      Crc32CChecksum firstChecksum = calculateCRC32C(stream, (int) s3Object.getSize());
+
+      S3URI s3URI =
+          s3Object.getObjectUri(this.getS3ExecutionContext().getConfiguration().getBaseUri());
+      S3AsyncClient s3Client = this.getS3ExecutionContext().getS3Client();
+
+      // Change the file content
+      s3Client
+          .putObject(
+              x -> x.bucket(s3URI.getBucket()).key(s3URI.getKey()),
+              AsyncRequestBody.fromBytes(generateRandomBytes((int) s3Object.getSize())))
+          .join();
+
+      // Create second stream
+      S3SeekableInputStream secondStream = s3AALClientStreamReader.createReadStream(s3Object);
+      Crc32CChecksum secondChecksum = calculateCRC32C(secondStream, (int) s3Object.getSize());
+
+      if (s3Object.getSize() < 8 * ONE_MB) {
+        // For small files, checksums should match as data should come from cache
+        assertChecksums(firstChecksum, secondChecksum);
+      } else {
+        // For large files, checksums should be different as second read gets new content
+        assertNotEquals(
+            firstChecksum.getChecksumBytes(),
+            secondChecksum.getChecksumBytes(),
+            "For large files, checksums should be different after file modification");
+      }
     }
   }
 
--- a/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
+++ b/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
@@ -44,6 +44,8 @@ public class PhysicalIOConfiguration {
   private static final long DEFAULT_BLOCK_READ_TIMEOUT = 30_000;
   private static final int DEFAULT_BLOCK_READ_RETRY_COUNT = 20;
   private static final int DEFAULT_MEMORY_CLEANUP_FREQUENCY_MILLISECONDS = 5000;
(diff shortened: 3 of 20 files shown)


input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
```diff
@@ -41,6 +41,8 @@ public class PhysicalIOConfiguration {
   private static final double DEFAULT_SEQUENTIAL_PREFETCH_SPEED = 1.0;
   private static final long DEFAULT_BLOCK_READ_TIMEOUT = 30_000;
   private static final int DEFAULT_BLOCK_READ_RETRY_COUNT = 20;
+  private static final boolean DEFAULT_SMALL_OBJECTS_PREFETCHING_ENABLED = true;
+  private static final long DEFAULT_SMALL_OBJECT_SIZE_THRESHOLD = 3 * ONE_MB;
```
ahmarsuhail, line 45: while you're here, think we can make this threshold 8MB?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManager.java
```diff
@@ -97,6 +104,62 @@ public BlockManager(
     this.ioPlanner = new IOPlanner(blockStore);
     this.rangeOptimiser = new RangeOptimiser(configuration);
     this.streamContext = streamContext;
+
+    initializeBlockManager();
```
ahmarsuhail, line 108: rename this to `maybePrefetchSmallObject` or similar. since you're not doing anything else in this initialisation

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManager.java
```diff
@@ -97,6 +104,62 @@ public BlockManager(
...
+          });
+    }
+  }
+
+  /**
+   * Checks if the current object qualifies as a small object based on configuration.
+   *
+   * @return true if the object should be treated as a small object
+   */
+  private boolean isSmallObject() {
```
ahmarsuhail, line 144: can we move this to a static utils class please. add a new class `AnalyticsAcceleratorUtils` or `AALUtils` or similar in the `software.amazon.s3.analyticsaccelerator.util` package and use that. Keeps this class cleaner!

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManager.java
```diff
@@ -97,6 +104,62 @@ public BlockManager(
...
+              LOG.warn(
+                  "Failed to prefetch small object for key: {}", objectKey.getS3URI().getKey(), e);
+              telemetry.measureStandard(
+                  () ->
+                      Operation.builder()
+                          .name("block.manager.small.object.prefetch.failure")
+                          .attribute(StreamAttributes.uri(this.objectKey.getS3URI()))
+                          .attribute(StreamAttributes.etag(this.objectKey.getEtag()))
+                          .build(),
+                  () -> {
```
ahmarsuhail, line 131: you don't need telemetry here. just a log at debug level is fine.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManager.java
```diff
@@ -97,6 +104,62 @@ public BlockManager(
...
+   */
+  private void prefetchSmallObject() throws IOException {
+    this.telemetry.measureStandard(
+        () ->
+            Operation.builder()
+                .name(OPERATION_SMALL_OBJECT_PREFETCH)
+                .attribute(StreamAttributes.uri(this.objectKey.getS3URI()))
+                .attribute(StreamAttributes.etag(this.objectKey.getEtag()))
+                .build(),
+        () -> makeRangeAvailable(0, metadata.getContentLength(), ReadMode.ASYNC));
```
ahmarsuhail, line 162: can you add a new mode to the ReadMode, "SMALL_OBJECT_ASYNC" or similar. and use that here. helps differentiate where exactly the request is coming from in access logs!

ahmarsuhail, line 162: instead of having a separate method here, think it's ok for you to move this logic up to your try block

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManagerTest.java
```diff
@@ -111,14 +112,14 @@ void testGetBlockIsEmpty() throws IOException {
...
-    BlockManager blockManager = getTestBlockManager(65 * ONE_KB);
+    BlockManager blockManager = getTestBlockManager(4 * ONE_MB);
 
     // When: have a 64KB block available from 0
     blockManager.makePositionAvailable(0, ReadMode.SYNC);
 
     // Then: 0 returns a block but 64KB + 1 byte returns no block
     assertTrue(blockManager.getBlock(0).isPresent());
-    assertFalse(blockManager.getBlock(64 * ONE_KB).isPresent());
+    assertFalse(blockManager.getBlock(4 * ONE_MB).isPresent());
```
ahmarsuhail, line 122: you shouldn't need to change this, when your file size is larger than your small object threshold, which in this case it is, this assertion will still hold true.

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManagerTest.java
```diff
@@ -288,4 +289,42 @@ private BlockManager getTestBlockManager(
...
+  @Test
+  void testSmallObjectPrefetching() throws IOException {
+    ObjectClient objectClient = mock(ObjectClient.class);
+    int smallObjectSize = 2 * ONE_MB; // Size less than default threshold (3MB)
+
+    BlockManager blockManager =
+        getTestBlockManager(objectClient, smallObjectSize, PhysicalIOConfiguration.DEFAULT);
+
+    // Verify that entire object was prefetched
+    ArgumentCaptor<GetRequest> requestCaptor = ArgumentCaptor.forClass(GetRequest.class);
```
ahmarsuhail, line 302: why do you need the timeout? 

can see other tests just do 

```
    ArgumentCaptor<GetRequest> requestCaptor = ArgumentCaptor.forClass(GetRequest.class);
    verify(objectClient).getObject(requestCaptor.capture(), any());
```

ahmarsuhail (changes requested):
looking mostly good! 

Some minor comments. Let's update default threshold to 8MB, let's add in an integration test, then we should be good to go.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManager.java
```diff
@@ -97,6 +104,62 @@ public BlockManager(
...
+   * asynchronously to avoid blocking the constructor.
+   */
+  private void initializeBlockManager() {
+    if (isSmallObject()) {
+      CompletableFuture.runAsync(
+          () -> {
+            try {
+              prefetchSmallObject();
+            } catch (IOException e) {
+              LOG.warn(
```
rajdchak, line 122: should this be a warn or error log?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManager.java
rajdchak: +1

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManager.java
rajdchak: do we actually need to measure this operation? this just calls makeRangeAvailable and we already measure the makeRangeAvailable operation.

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/IntegrationTestBase.java
```diff
@@ -322,6 +324,59 @@ protected void testAALReadConcurrency(
...
+   * smaller than 8MB, the entire object should be prefetched into cache on first read. For objects
+   * larger than 8MB, data should be fetched from S3 as needed.
+   *
+   * @param s3ClientKind the type of S3 client to use
+   * @param s3Object the S3 object to test with
+   * @param streamReadPatternKind the pattern to use for reading the stream
+   * @param AALInputStreamConfigurationKind the configuration for the input stream
+   * @throws IOException if an I/O error occurs during the test
+   */
+  protected void testSmallObjectPrefetching(
```
ahmarsuhail, line 338: thanks! this is a good workaround test :) can I ask for a TODO here. 

TODO: Update test to assert on GET request metrics once we have them. 

Esentially we are starting to add stats now, what you want to check here is that for small files, only a single GET was made. for everything else.. we made multiple GETs.  Going forward, we really don't want checksum behaviour to assert on anything, so we will need to update this test!

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
```diff
@@ -162,7 +185,10 @@ private PhysicalIOConfiguration(
       double sequentialPrefetchBase,
       double sequentialPrefetchSpeed,
       long blockReadTimeout,
-      int blockReadRetryCount) {
+      int blockReadRetryCount,
+      boolean smallObjectsPrefetchingEnabled,
+      long smallObjectSizeThreshold) {
+
     Preconditions.checkArgument(blobStoreCapacity > 0, "`blobStoreCapacity` must be positive");
```
ahmarsuhail, line 192: think you need a pre-condition check here for small.object.size.threshold should be > 0

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
```diff
@@ -132,6 +149,10 @@ public static PhysicalIOConfiguration fromConfiguration(ConnectorConfiguration c
         .blockReadTimeout(configuration.getLong(BLOCK_READ_TIMEOUT_KEY, DEFAULT_BLOCK_READ_TIMEOUT))
         .blockReadRetryCount(
             configuration.getInt(BLOCK_READ_RETRY_COUNT_KEY, DEFAULT_BLOCK_READ_RETRY_COUNT))
+        .smallObjectsPrefetchingEnabled(
+            configuration.getBoolean(SMALL_OBJECTS_PREFETCHING_ENABLED_KEY, true))
```
rajdchak, line 153: why not set this to DEFAULT_SMALL_OBJECTS_PREFETCHING_ENABLED?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
```diff
@@ -132,6 +149,10 @@ public static PhysicalIOConfiguration fromConfiguration(ConnectorConfiguration c
...
+        .smallObjectSizeThreshold(
+            configuration.getLong(SMALL_OBJECT_SIZE_THRESHOLD_KEY, 8 * ONE_MB))
```
rajdchak, line 155: same as above, why not DEFAULT_SMALL_OBJECT_SIZE_THRESHOLD?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
rajdchak: +1

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
```diff
@@ -41,6 +41,8 @@ public class PhysicalIOConfiguration {
   private static final double DEFAULT_SEQUENTIAL_PREFETCH_SPEED = 1.0;
   private static final long DEFAULT_BLOCK_READ_TIMEOUT = 30_000;
   private static final int DEFAULT_BLOCK_READ_RETRY_COUNT = 20;
+  private static final boolean DEFAULT_SMALL_OBJECTS_PREFETCHING_ENABLED = true;
+  private static final long DEFAULT_SMALL_OBJECT_SIZE_THRESHOLD = 8 * ONE_MB;
```
rajdchak, line 46: Nit: Here and in all places can we rename this to something like small_object_prefetch_size?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/util/ObjectSize.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package software.amazon.s3.analyticsaccelerator.util;
+
+import software.amazon.s3.analyticsaccelerator.io.physical.PhysicalIOConfiguration;
+
+/**
+ * Utility class for object size-related operations and determinations. Provides methods to classify
+ * objects based on their size and configuration settings.
+ */
+public class ObjectSize {
```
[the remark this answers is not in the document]
rajdchak: +1

input-stream/src/testFixtures/java/software/amazon/s3/analyticsaccelerator/access/S3Object.java
```diff
@@ -95,12 +95,16 @@ public static List<S3Object> smallObjects() {
...
    *
    * @return list of small binary objects
    */
   public static List<S3Object> smallBinaryObjects() {
-    return filter(o -> o.size < 50 * SizeConstants.ONE_MB_IN_BYTES && o.getName().endsWith(".bin"));
+    return filter(
+        o ->
+            o.size >= 8 * SizeConstants.ONE_MB_IN_BYTES
+                && o.size < 50 * SizeConstants.ONE_MB_IN_BYTES
+                && o.getName().endsWith(".bin"));
```
rajdchak, line 107: should we make this better by having these as static variables defined, like SMALL_BINARY_OBJECTS_UPPER_LIMIT and LOWER_LIMIT?

doc/CONFIGURATION.md
```diff
@@ -60,16 +58,17 @@ Options under `<CONNECTOR_PREFIX>.logicalio.`
...
+| `metadatastore.capacity`   | `50`    | Capacity of the metadata store                 |
+| `blocksizebytes`           | `8MB`   | Size of blocks for data transfer               |
+| `readaheadbytes`           | `64KB`  | Number of bytes to read ahead                  |
+| `maxrangesizebytes`        | `8MB`   | Maximum size of range requests                 |
+| `partsizebytes`            | `8MB`   | Size of individual parts for transfer          |
+| `sequentialprefetch.base`  | `2.0`   | Base factor for sequential prefetch sizing     |
+| `sequentialprefetch.speed` | `1.0`   | Speed factor for sequential prefetch growth    |
+|  `small.objects.prefetching.enabled` | `true`  | Controls prefetching for small objects |
+|`small.object.size.threshold` | `8MB`   | Size of page index to prefetch for large files |
 ## Telemetry Configuration
```
rajdchak, line 72: is this for large file?

doc/CONFIGURATION.md
```diff
@@ -60,16 +58,17 @@ Options under `<CONNECTOR_PREFIX>.logicalio.`
...
+|----------------------------|---------|------------------------------------------------|
+| `metadatastore.capacity`   | `50`    | Capacity of the metadata store                 |
+| `blocksizebytes`           | `8MB`   | Size of blocks for data transfer               |
+| `readaheadbytes`           | `64KB`  | Number of bytes to read ahead                  |
+| `maxrangesizebytes`        | `8MB`   | Maximum size of range requests                 |
+| `partsizebytes`            | `8MB`   | Size of individual parts for transfer          |
+| `sequentialprefetch.base`  | `2.0`   | Base factor for sequential prefetch sizing     |
+| `sequentialprefetch.speed` | `1.0`   | Speed factor for sequential prefetch growth    |
+|  `small.objects.prefetching.enabled` | `true`  | Controls prefetching for small objects |
+|`small.object.size.threshold` | `8MB`   | Size of page index to prefetch for large files |
```
rajdchak, line 71: is this for large file?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
ahmarsuhail: +1, let's update the name of this default pls

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManagerTest.java
```diff
@@ -117,14 +117,14 @@ void testGetBlockIsEmpty() throws IOException {
   @Test
   void testGetBlockReturnsAvailableBlock() throws IOException {
     // Given
-    BlockManager blockManager = getTestBlockManager(65 * ONE_KB);
+    BlockManager blockManager = getTestBlockManager(9 * ONE_MB);
```
ahmarsuhail, line 120: why are you changing this, this is a test for readAhead behaviour. Instead, you need to turn off small object prefetching for this test.

input-stream/src/testFixtures/java/software/amazon/s3/analyticsaccelerator/access/S3Object.java
ahmarsuhail: why ends with with .bin? 

think you only need a single condition here : o.size <= SMALL_OBJECTS_LOWER_LIMIT

ahmarsuhail: oh is it .bin cause the others will use the other logicalIO and get prefetched anyway

ahmarsuhail (approved):
+1, LGTM

rajdchak (approved):

PR #266, opened 2025-05-09
ElastiCache cache for parquet metadata

petergalati:
## Description of change

Implements integration with Valkey ElastiCache in order to cache repeated calls to S3 for parquet file metadata.

TODO: Add metrics to track cache hits / misses and actual savings compared to S3.

No

No

Unit tests and by running internal benchmarks

---

### DIFF
--- a/input-stream/build.gradle.kts
+++ b/input-stream/build.gradle.kts
@@ -72,6 +72,7 @@ dependencies {
     implementation(project(":common"))
     implementation(libs.parquet.format)
     implementation(libs.slf4j.api)
+    implementation("io.valkey:valkey-java:5.3.0")
 
     jmhImplementation(libs.s3)
     jmhImplementation(libs.s3.transfer.manager)
@@ -148,8 +149,6 @@ val shadowJar = tasks.withType<ShadowJar> {
 }
 
 val refTest = task<Test>("referenceTest") {
-    maxHeapSize = "2g"
-    jvmArgs = listOf("-Xms1g", "-Xmx2g")
     description = "Runs reference tests."
     group = "verification"
 
@@ -375,10 +374,10 @@ publishing {
         maven {
             name = "sonatype"
             url = uri("https://aws.oss.sonatype.org/service/local/staging/deploy/maven2/")
-           credentials {
-               username = findProperty("mavenUsername") as String? ?: ""
-               password = findProperty("mavenPassword") as String? ?: ""
-           }
+            credentials {
+                username = findProperty("mavenUsername") as String? ?: ""
+                password = findProperty("mavenPassword") as String? ?: ""
+            }
         }
     }
 }
--- a/input-stream/build.gradle.kts
+++ b/input-stream/build.gradle.kts
@@ -10,7 +10,7 @@ import com.github.jk1.license.render.TextReportRenderer
 
 val group = "software.amazon.s3.analyticsaccelerator"
 val artefact = "analyticsaccelerator-s3"
-val currentVersionNumber = "1.0.0"
+val currentVersionNumber = "2.0.0"
 
 val isSnapshot = findProperty("snapshotBuild") == "true"
 val currentVersion = if (isSnapshot) "SNAPSHOT" else currentVersionNumber;
--- a/input-stream/build.gradle.kts
+++ b/input-stream/build.gradle.kts
@@ -10,7 +10,7 @@ import com.github.jk1.license.render.TextReportRenderer
 
 val group = "software.amazon.s3.analyticsaccelerator"
 val artefact = "analyticsaccelerator-s3"
-val currentVersionNumber = "2.0.0"
+val currentVersionNumber = "1.0.0"
 
 val isSnapshot = findProperty("snapshotBuild") == "true"
 val currentVersion = if (isSnapshot) "SNAPSHOT" else currentVersionNumber;
--- a/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStreamFactory.java
+++ b/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStreamFactory.java
@@ -87,9 +87,9 @@ public class S3SeekableInputStreamFactory implements AutoCloseable {
       this.cache =
           new ValkeyCacheImpl(configuration.getPhysicalIOConfiguration().getCacheEndpoint());
 
-      this.executorService = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors() * 10);
-
       LOG.info("Cache successfully instantiated");
+
+      this.executorService = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors() * 10);
 
     } else {
       LOG.info("Cache disabled");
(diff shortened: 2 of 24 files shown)


input-stream/build.gradle.kts
```diff
@@ -10,7 +10,7 @@ import com.github.jk1.license.render.TextReportRenderer
 
 val group = "software.amazon.s3.analyticsaccelerator"
 val artefact = "analyticsaccelerator-s3"
-val currentVersionNumber = "1.0.0"
+val currentVersionNumber = "2.0.0"
```
ahmarsuhail, line 13: revert to 1.0.0 pls

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStreamFactory.java
```diff
@@ -180,5 +193,8 @@ public void close() throws IOException {
     this.objectMetadataStore.close();
     this.objectBlobStore.close();
     this.telemetry.close();
+    if (cache instanceof ValkeyCacheImpl) {
```
ahmarsuhail, line 196: you can just do a null check here? and then call cache.close(). why do you need to do `((ValkeyCacheImpl) cache).close();`

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/Cache.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package software.amazon.s3.analyticsaccelerator.io.physical;
+
+/** An interface defining how the ElastiCache cache should behave */
```
ahmarsuhail, line 18: needs a better java doc, something like "Cache interface, to be used for staring data in a shared cache that may be accessed in by any node for the duration of a workload"

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
```diff
@@ -330,4 +416,24 @@ public void close() {
...
+
+  /**
+   * Checks if the current block is a tail block.
+   *
+   * @return true if the current block is a tail block, false otherwise
+   */
+  private boolean isTailMetadata() {
+    /** hardcoded right now such that ONE_MB matches DEFAULT_PREFETCH_LARGE_FILE_METADATA_SIZE */
+    //    TODO: find a way to adjust this dynamically instead of just using ONE_MB (if necessary)
+    return start + 8 * ONE_MB + 8 * ONE_MB >= contentLength;
```
ahmarsuhail, line 428: i think we can annotate footer reads from the logicalIO, and then you can use that here. Add a range type field to the Range.java class, and then In `ParquetPrefetchTailTask` set it to "footer". and then use that here

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/impl/ValkeyCacheImpl.java
```diff
@@ -0,0 +1,77 @@
...
+   * @param endpoint the ElastiCache endpoint of the serverless Valkey cache
+   */
+  public ValkeyCacheImpl(String endpoint) {
+
+    ConnectionPoolConfig config = new ConnectionPoolConfig();
+    config.setMaxTotal(64);
+    config.setMaxIdle(64);
+    config.setMinIdle(32);
+
+    JedisClientConfig clientConfig = DefaultJedisClientConfig.builder().ssl(true).build();
```
ahmarsuhail, line 42: why does ssl need to be true?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/Cache.java
```diff
@@ -0,0 +1,39 @@
...
+public interface Cache {
+
+  /**
+   * Fetches the value from ElastiCache given a key
+   *
+   * @param key the key to fetch from ElastiCache
+   * @return the value associated with the key in ElastiCache
+   */
+  byte[] get(byte[] key);
```
ahmarsuhail, line 27: does the get key really need to be a byte array?

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockTest.java
```diff
@@ -339,4 +333,74 @@ void testClose() {
...
+            true,
+            mockCache,
+            null);
+
+    byte[] buffer = new byte[TEST_DATA.length()];
+    block.read(buffer, 0, buffer.length, 0);
+
+    assertEquals(TEST_DATA, new String(buffer, StandardCharsets.UTF_8));
+
+    verify(mockCache).get(any());
```
ahmarsuhail, line 367: you should verify how many times get was called, think you can do

verify(mockCache, 1).get(any(byte[].class);

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockTest.java
```diff
@@ -339,4 +333,74 @@ void testClose() {
...
```
ahmarsuhail, line 402: same here, you can improve your verify statements,

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockTest.java
```diff
@@ -339,4 +333,74 @@ void testClose() {
     block.close();
     block.close();
   }
+
+  @Test
+  void testCacheHitForTailMetadata() throws IOException {
```
ahmarsuhail, line 338: we need to be able to verify that you're only caching the metadata

ahmarsuhail (changes requested):
thanks! mostly looking good, some minor comments.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/Cache.java
petergalati: could make it string instead, do you reckon it would be better that way?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/impl/ValkeyCacheImpl.java
petergalati: true, it likely doesn't need to since we're making all requests within the VPC. will look into if this allows for any performance improvements.

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockTest.java
petergalati: Sure, will do that in a separate test, to verify the generateCacheKey() function is working as intended.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/impl/ValkeyCacheImpl.java
petergalati: Ah actually I think enabling ssl is necessary since it may be a requirement to establish connection to elasticache. when set to false i get 

```
Error during cache testing: Could not initialize cluster slots cache.
io.valkey.exceptions.JedisClusterOperationException: Could not initialize cluster slots cache.
        at io.valkey.providers.ClusterConnectionProvider.initializeSlotsCache(ClusterConnectionProvider.java:68)
        at io.valkey.providers.ClusterConnectionProvider.<init>(ClusterConnectionProvider.java:35)
        at io.valkey.JedisCluster.<init>(JedisCluster.java:210)
        ...
```

also on elasticache dashboard the template client config has `.ssl(true) `

PR #268, opened 2025-05-09
Eliminate repeated linear searches in BlockManager

ozkoca:
## Description of change
Currently, BlockManager performs three linear searches in BlockStore when handling a sync read that continues from a previous block:
i. To check if the read is sequential
ii. Again inside getGeneration to verify sequentiality
iii. To fetch the generation of the previous block

This PR reduces the overhead by collapsing these into a single lookup: it retrieves the previous block once and determines both sequentiality and generation based on its existence.

#### Relevant issues
N/A

No

No

Executed unit tests

- No

---


input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/prefetcher/SequentialPatternDetector.java
```diff
@@ -1,70 +0,0 @@
...
-import software.amazon.s3.analyticsaccelerator.io.physical.data.BlockStore;
-
-/**
- * Class capable of detecting sequential read patterns.
- *
- * <p>The SequentialPatternDetector depends on the BlockStore which it is capable of 'inspecting'.
- * For now, pattern detection is very simple: if the previous byte is already stored by a block,
- * then we conclude that the byte is requested by a sequential read. In the future we should
- * probably extend this to work with some kind of tolerance or radius (i.e., look back more than a
- * single byte).
```
rajdchak, line 29: I think we wanted to extend this class in future. I don't think we should remove it and take the logic to BlockManager.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManager.java
```diff
@@ -194,14 +191,15 @@ public synchronized void makeRangeAvailable(long pos, long len, ReadMode readMod
     // TODO: Improve readModes, as tracked in
     // https://github.com/awslabs/analytics-accelerator-s3/issues/195
     final long generation;
-    if (readMode != ReadMode.ASYNC && patternDetector.isSequentialRead(pos)) {
-      generation = patternDetector.getGeneration(pos);
+    if (readMode == ReadMode.ASYNC || pos == 0) {
```
vaibhav5140, line 194: This will trigger prefetching for any non-ASYNC, non-zero position read, even if it's not actually sequential 

Example:
Read pattern:
1. Read 4KB at position 0
2. Read 4KB at position 500KB (random jump)
3. Read 4KB at position 250KB (another random jump)

Old behavior (correct):
- All positions: No prefetch (not sequential)

New behavior:
- Position 0: No prefetch (pos == 0)
- Position 500KB: Triggers prefetch (neither ASYNC nor pos 0)
- Position 250KB: Triggers prefetch (neither ASYNC nor pos 0)

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManager.java
```diff
@@ -194,14 +191,15 @@ public synchronized void makeRangeAvailable(long pos, long len, ReadMode readMod
...
+      generation = 0;
+    } else {
+      // Check if it is sequential pattern
+      generation = blockStore.getBlock(pos - 1).map(block -> block.getGeneration() + 1).orElse(0L);
```
ahmarsuhail, line 198: nice!

ahmarsuhail (changes requested):
looks good, think we're losing out on some test coverage though?

do we still have tests that will verify if a block with pos-1 exists, then it's generation will be increased?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/prefetcher/SequentialPatternDetector.java
ozkoca: @rajdchak in its current form, it's introducing a performance bottleneck, which is why I opted to simplify it and move the logic into BlockManager. My thinking is that we can always reintroduce a well-designed, extensible version of the class later if and when we need it

fuatbasik: keeping SequentialPatternDetector around might be helpful. I can easily see how we are going to extend these patterns to cover nearly sequential reads (sequential read with wholes).

PR #281, opened 2025-06-03
SSE_C changes

rajdchak:
## Description of change

Support passing of sse_c customer key to pass this to s3 for encryption/decryption. 

S3A currently has this customer key as Optional String https://github.com/apache/hadoop/blob/trunk/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/auth/delegation/EncryptionSecretOperations.java#L41 while Iceberg has this key as a String https://github.com/apache/iceberg/blob/f9cc62eb0d98e360b452a3ab8fdc6efdc4969f6e/aws/src/main/java/org/apache/iceberg/aws/s3/S3FileIOProperties.java#L499. So decided to accept this key as Optional String.

---

### DIFF
--- a/common/src/test/java/software/amazon/s3/analyticsaccelerator/util/OpenStreamInformationTest.java
+++ b/common/src/test/java/software/amazon/s3/analyticsaccelerator/util/OpenStreamInformationTest.java
@@ -28,6 +28,14 @@ import software.amazon.s3.analyticsaccelerator.request.StreamAuditContext;
 
 public class OpenStreamInformationTest {
 
+  private static final String CUSTOMER_KEY = "32-bytes-long-key-for-testing-123";
+
+  /**
+   * To generate the base64 encoded md5 value for a customer key use the cli command echo -n
+   * "customer_key" | base64 | base64 -d | openssl md5 -binary | base64
+   */
+  private static final String EXPECTED_BASE64_MD5 = "R+k8pqEVUmkxDfaH5MqIdw==";
+
   @Test
   public void testDefaultInstance() {
     OpenStreamInformation info = OpenStreamInformation.DEFAULT;
@@ -45,8 +53,7 @@ public class OpenStreamInformationTest {
     ObjectMetadata mockMetadata = Mockito.mock(ObjectMetadata.class);
     InputPolicy mockPolicy = Mockito.mock(InputPolicy.class);
     String base64Key =
-        Base64.getEncoder()
-            .encodeToString("32-bytes-long-key-for-testing-123".getBytes(StandardCharsets.UTF_8));
+        Base64.getEncoder().encodeToString(CUSTOMER_KEY.getBytes(StandardCharsets.UTF_8));
     EncryptionSecrets secrets =
         EncryptionSecrets.builder().sseCustomerKey(Optional.of(base64Key)).build();
 
@@ -66,7 +73,7 @@ public class OpenStreamInformationTest {
         info.getEncryptionSecrets().getSsecCustomerKey().get(),
         "Customer key should match");
     assertNotNull(info.getEncryptionSecrets().getSsecCustomerKeyMd5(), "MD5 should not be null");
-    assertEquals("R+k8pqEVUmkxDfaH5MqIdw==", info.getEncryptionSecrets().getSsecCustomerKeyMd5());
+    assertEquals(EXPECTED_BASE64_MD5, info.getEncryptionSecrets().getSsecCustomerKeyMd5());
   }
 
   @Test
@@ -131,8 +138,7 @@ public class OpenStreamInformationTest {
   public void testBuilderWithEncryptionSecrets() {
     // Create a sample base64 encoded key
     String base64Key =
-        Base64.getEncoder()
-            .encodeToString("32-bytes-long-key-for-testing-123".getBytes(StandardCharsets.UTF_8));
+        Base64.getEncoder().encodeToString(CUSTOMER_KEY.getBytes(StandardCharsets.UTF_8));
     EncryptionSecrets secrets =
         EncryptionSecrets.builder().sseCustomerKey(Optional.of(base64Key)).build();
 
@@ -147,7 +153,7 @@ public class OpenStreamInformationTest {
         info.getEncryptionSecrets().getSsecCustomerKey().get(),
         "Customer key should match");
     assertNotNull(info.getEncryptionSecrets().getSsecCustomerKeyMd5(), "MD5 should not be null");
-    assertEquals("R+k8pqEVUmkxDfaH5MqIdw==", info.getEncryptionSecrets().getSsecCustomerKeyMd5());
+    assertEquals(EXPECTED_BASE64_MD5, info.getEncryptionSecrets().getSsecCustomerKeyMd5());
   }
 
   @Test


object-client/src/main/java/software/amazon/s3/analyticsaccelerator/S3SdkObjectClient.java
```diff
@@ -121,6 +129,15 @@ public CompletableFuture<ObjectMetadata> headObject(HeadRequest headRequest) {
             .putHeader(HEADER_USER_AGENT, this.userAgent.getUserAgent())
             .build());
 
+    if (openStreamInformation != null
```
ahmarsuhail, line 132: we probably don't need the null checks..

common/build.gradle.kts
```diff
@@ -10,6 +10,7 @@ plugins {
 
 dependencies {
     implementation(libs.slf4j.api)
+    implementation(libs.sdk.core)
```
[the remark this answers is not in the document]
fuatbasik: Yes, only part depending on S3 SDK of AAL is S3SDKObjectClient. This is why in Common we have the request interfaces but not the actually request logic. We design AAL to be object-store agnostic. AAL should be used by store implementing S3 interface.

common/src/main/java/software/amazon/s3/analyticsaccelerator/request/EncryptionSecrets.java
```diff
@@ -0,0 +1,59 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package software.amazon.s3.analyticsaccelerator.request;
+
+import java.util.Base64;
+import java.util.Optional;
+import lombok.Builder;
+import lombok.Getter;
+import software.amazon.awssdk.utils.Md5Utils;
```
fuatbasik, line 22: Lets avoid bringing in SDK just to compute MD5. AFAIK java.security would already have helpers for this.

gradle/libs.versions.toml
```diff
@@ -4,6 +4,7 @@ mockito = "4.11.0"
 spotless = "6.13.0"
 spotbugs = "6.0.20"
 s3 = "2.25.31"
+sdk-core = "2.25.31"
```
fuatbasik, line 7: i am not sure we need a new dependency to SDK for SSE-C. Unless we need it for 100% lets avoid it.

input-stream/src/testFixtures/java/software/amazon/s3/analyticsaccelerator/access/S3Object.java
```diff
@@ -52,7 +52,9 @@ public enum S3Object {
   CSV_20MB(
       "sequential-20mb.csv", 20 * SizeConstants.ONE_MB_IN_BYTES, S3ObjectKind.RANDOM_SEQUENTIAL),
   TXT_16MB(
-      "sequential-16mb.txt", 16 * SizeConstants.ONE_MB_IN_BYTES, S3ObjectKind.RANDOM_SEQUENTIAL);
+      "sequential-16mb.txt", 16 * SizeConstants.ONE_MB_IN_BYTES, S3ObjectKind.RANDOM_SEQUENTIAL),
+  RANDOM_SSEC_ENCRYPTED_1MB(
```
fuatbasik, line 56: we need Parquet objects with SSE-C enabled for this testing. We should confirm LogicalIO works with such objects.

fuatbasik (commented):
Thanks @rajdchak . My major comment is around dependencies. We should not depend sdk.core for this change. We also should not add any S3 dependency to common.

input-stream/src/testFixtures/java/software/amazon/s3/analyticsaccelerator/access/S3Object.java
rajdchak: yes will add more tests, as of today we don't have any parquet objects for integration tests right?

common/src/main/java/software/amazon/s3/analyticsaccelerator/request/EncryptionSecrets.java
```diff
@@ -0,0 +1,69 @@
...
+  @Builder
+  public EncryptionSecrets(Optional<String> sseCustomerKey) {
+    this.ssecCustomerKey = sseCustomerKey;
+    this.ssecCustomerKeyMd5 =
+        sseCustomerKey
+            .map(
+                key -> {
+                  try {
+                    MessageDigest md = MessageDigest.getInstance("MD5");
+                    return Base64.getEncoder()
```
ahmarsuhail, line 61: why are we encoding to base64 again? think all we need is the md5. 

`Md5Utils.md5AsBase64(Base64.getDecoder().decode(base64customerKey))` to get the Md5. whey are we encoding it again? 

and what is throwing the `NoSuchAlgorithmException?` ? asking as in S3A, I see that all we do is:

```
      builder.sseCustomerAlgorithm(ServerSideEncryption.AES256.name())
          .sseCustomerKey(base64customerKey)
          .sseCustomerKeyMD5(Md5Utils.md5AsBase64(Base64.getDecoder().decode(base64customerKey)));
```

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/SSECEncryptionTest.java
```diff
@@ -0,0 +1,51 @@
...
+      S3ClientKind s3ClientKind,
+      S3Object s3Object,
+      StreamReadPatternKind streamReadPattern,
+      AALInputStreamConfigurationKind configuration)
+      throws IOException {
+    testReadPatternUsingSSECEncryption(s3ClientKind, s3Object, streamReadPattern, configuration);
+  }
+
+  static Stream<Arguments> encryptedReads() {
+    List<S3Object> readVectoredObjects = new ArrayList<>();
```
ahmarsuhail, line 41: nit: rename variable `readVectoredObjects`

common/src/main/java/software/amazon/s3/analyticsaccelerator/request/EncryptionSecrets.java
rajdchak: We need md5AsBase64 so we need to encode

common/src/test/java/software/amazon/s3/analyticsaccelerator/util/OpenStreamInformationTest.java
```diff
@@ -32,24 +36,36 @@ public void testDefaultInstance() {
...
             .build();
 
     assertSame(mockContext, info.getStreamAuditContext(), "StreamContext should match");
     assertSame(mockMetadata, info.getObjectMetadata(), "ObjectMetadata should match");
     assertSame(mockPolicy, info.getInputPolicy(), "InputPolicy should match");
+    assertEquals(
+        base64Key,
+        info.getEncryptionSecrets().getSsecCustomerKey().get(),
+        "Customer key should match");
+    assertNotNull(info.getEncryptionSecrets().getSsecCustomerKeyMd5(), "MD5 should not be null");
```
ahmarsuhail, line 68: you should assert on the md5 value of this. if you calculate what you expect the base64 md5 of `32-bytes-long-key-for-testing-123` to be, this should be equal

common/src/test/java/software/amazon/s3/analyticsaccelerator/util/OpenStreamInformationTest.java
```diff
@@ -103,4 +119,47 @@ public void testNullFields() {
...
+
+    assertNotNull(info.getEncryptionSecrets(), "EncryptionSecrets should not be null");
+    assertTrue(
+        info.getEncryptionSecrets().getSsecCustomerKey().isPresent(),
+        "Customer key should be present");
+    assertEquals(
+        base64Key,
+        info.getEncryptionSecrets().getSsecCustomerKey().get(),
+        "Customer key should match");
+    assertNotNull(info.getEncryptionSecrets().getSsecCustomerKeyMd5(), "MD5 should not be null");
```
ahmarsuhail, line 148: same here, assert on the md5 value

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/IntegrationTestBase.java
```diff
@@ -99,7 +101,48 @@ protected void testAndCompareStreamReadPattern(
...
+        streamReadPattern,
+        AALInputStreamConfigurationKind,
+        Optional.of(aalChecksum),
+        OpenStreamInformation.DEFAULT);
+
+    // Assert checksums
+    assertChecksums(directChecksum, aalChecksum);
+  }
+
+  protected void testReadPatternUsingSSECEncryption(
```
ahmarsuhail, line 124: Can you move this into `SSECEncryptionTest` ?

I think we've been getting this wrong, and this leads to the ITests becoming a bit of a mess. This is specific to a feature, so this go into `SSECEncryptionTest`. 

I'd actually never written an ITest in AAL until very recently so had no idea how this was working and missed this in my previous reviews.

input-stream/src/testFixtures/java/software/amazon/s3/analyticsaccelerator/access/S3AsyncClientStreamReader.java
```diff
@@ -54,26 +56,40 @@ public S3AsyncClientStreamReader(
...
+          GetObjectRequest.builder()
+              .bucket(s3URI.getBucket())
+              .key(s3URI.getKey())
+              .range(
+                  String.format(
+                      "bytes=%s-%s",
+                      streamRead.getStart(), streamRead.getStart() + streamRead.getLength() - 1));
+
+      // Add encryption parameters if present
+      if (openStreamInformation.getEncryptionSecrets() != null
```
ahmarsuhail, line 77: can you move this to a separate function, `addEncryptionParameters(GetObjectBuilder)`, so to keep this function clearer

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/SSECEncryptionTest.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.stream.Stream;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+public class SSECEncryptionTest extends IntegrationTestBase {
+
+  @ParameterizedTest
+  @MethodSource("encryptedReads")
+  void testEncryptedReads(
```
ahmarsuhail, line 31: can add some more tests here:

1/ read with a key different to one that the object was encrypted with and expect an accessdeniedException.

look at https://github.com/apache/hadoop/blob/trunk/hadoop-tools/hadoop-aws/src/test/java/org/apache/hadoop/fs/s3a/ITestS3AEncryptionSSEC.java for other test cases

ahmarsuhail (changes requested):
thanks, mostly looks good, have some minor comments

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/SSECEncryptionTest.java
rajdchak: yes I will be adding more tests, not yet done.

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/SSECEncryptionTest.java
```diff
@@ -0,0 +1,172 @@
...
+      S3Object s3Object,
+      StreamReadPatternKind streamReadPattern,
+      AALInputStreamConfigurationKind configuration)
+      throws IOException {
+    testReadPatternUsingSSECEncryption(
+        s3ClientKind,
+        s3Object,
+        streamReadPattern,
+        configuration,
+        "AO8XKQXJgtIS9G+IrSWZ2eSNW1yJlvqElVoVcNlvDqE=");
```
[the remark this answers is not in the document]
fuatbasik: We probably need to find a way to completely delete this PR from public space, delete any objects written with this encryption key.

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/SSECEncryptionTest.java
```diff
@@ -0,0 +1,172 @@
...
+  @MethodSource("encryptedReadsWithWrongKey")
+  void testEncryptedReadsWithWrongKey(
+      S3ClientKind s3ClientKind,
+      S3Object s3Object,
+      StreamReadPatternKind streamReadPattern,
+      AALInputStreamConfigurationKind configuration) {
+
+    CompletionException exception =
+        assertThrows(
+            CompletionException.class,
```
fuatbasik, line 81: is this the exception we like to throw? If we are using different clients, will the exception change?

rajdchak: My bad, we throw IOException, but somehow this exception is coming from the readWithoutAAL, will update this test.

ahmarsuhail (approved):
+1, LGTM. 

test file generation to be addressed in a follow up.

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/IntegrationTestBase.java
```diff
@@ -140,7 +146,8 @@ protected void testChangingEtagMidStream(
       S3URI s3URI =
           s3Object.getObjectUri(this.getS3ExecutionContext().getConfiguration().getBaseUri());
       S3AsyncClient s3Client = this.getS3ExecutionContext().getS3Client();
-      S3SeekableInputStream stream = s3AALClientStreamReader.createReadStream(s3Object);
+      S3SeekableInputStream stream =
+          s3AALClientStreamReader.createReadStream(s3Object, OpenStreamInformation.DEFAULT);
```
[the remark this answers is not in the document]
fuatbasik: This is a nit btw but probably a better way to implement this pattern.

input-stream/src/testFixtures/java/software/amazon/s3/analyticsaccelerator/access/S3StreamReaderBase.java
```diff
@@ -46,11 +47,13 @@ protected S3StreamReaderBase(@NonNull S3URI baseUri, int bufferSize) {
    * @param s3Object S3 Object to read
    * @param streamReadPattern Stream read pattern
    * @param checksum optional checksum, to update
+   * @param openStreamInformation contains the open stream information
    */
   public abstract void readPattern(
```
fuatbasik, line 52: do we want to change existing method signature or add a new method?

object-client/src/main/java/software/amazon/s3/analyticsaccelerator/S3SdkObjectClient.java
```diff
@@ -132,6 +133,16 @@ public CompletableFuture<ObjectMetadata> headObject(
 
     builder.overrideConfiguration(requestOverrideConfigurationBuilder.build());
 
+    if (openStreamInformation.getEncryptionSecrets() != null
```
fuatbasik, line 136: do we want to move this to a method that accepts a requestBuilder and adds encryption secrets if they are available. We are already duplicating these lines with 2 S3 methods and going forward, adding PUT etc. we will definitely need that.

ahmarsuhail: I've tried to do this before for S3A, annoying the actual methods we care about, like [sseCustomerKey](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/model/HeadObjectRequest.Builder.html#sseCustomerKey(java.lang.String)) are implemented on the individual request builders, like HeadRequestBuilder, and not the base class https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/model/S3Request.Builder.html, so it's not possible to do this. or at least i couldn't figure out how last time i tried :)

PR #283, opened 2025-06-04
Pass down openstreaminfo instead of streamcontext

rajdchak:

---

### DIFF
--- a/input-stream/src/referenceTest/java/software/amazon/s3/analyticsaccelerator/property/InMemoryS3SeekableInputStream.java
+++ b/input-stream/src/referenceTest/java/software/amazon/s3/analyticsaccelerator/property/InMemoryS3SeekableInputStream.java
@@ -81,20 +81,10 @@ public class InMemoryS3SeekableInputStream extends SeekableInputStream {
     }
 
     @Override
-    public CompletableFuture<ObjectMetadata> headObject(HeadRequest headRequest) {
-      return headObject(headRequest, null);
-    }
-
-    @Override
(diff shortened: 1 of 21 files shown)


input-stream/src/referenceTest/java/software/amazon/s3/analyticsaccelerator/property/InMemoryS3SeekableInputStream.java
```diff
@@ -81,6 +82,12 @@ public InMemoryObjectClient(int size) {
 
     @Override
     public CompletableFuture<ObjectMetadata> headObject(HeadRequest headRequest) {
+      return headObject(headRequest, null);
```
ahmarsuhail, line 85: why do we need headObject(HeadRequest headRequest)? 

think just having ```headObject(
        HeadRequest headRequest, OpenStreamInformation openStreamInformation)``` is enough. same for GetObject.

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/io/physical/data/IOPlannerTest.java
```diff
@@ -37,7 +37,7 @@
 @SuppressWarnings("unchecked")
 public class IOPlannerTest {
   private static final S3URI TEST_URI = S3URI.of("foo", "bar");
-  private static final String ETAG = "RandomString";
+  private static final String ETAG = "RANDOM";
```
ahmarsuhail, line 40: why do we need to change this?

PR #284, opened 2025-06-05
Adds test cases for readVectored()

ahmarsuhail:
## Description of change

Adds in more test cases for readVectored()

---


input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/ReadVectoredTest.java
```diff
@@ -41,6 +60,139 @@ void testVectoredReads(
...
+
+      assertThrows(
+          NullPointerException.class,
+          () ->
+              s3SeekableInputStream.readVectored(objectRanges, allocate, LOG_BYTE_BUFFER_RELEASED));
+    }
+  }
+
+  @Test
+  void testNullRangeList() throws IOException {
```
SanjayMarreddi, line 128: Should we make `testNullRangeList` a unit test at `S3SeekableInputStream` level as it just checks the arguments at the same level?

SanjayMarreddi (commented):
Thanks a lot for the PR!

I just want to see your thoughts on if we can move some of these tests to unit-tests as they test a single logical unit of functionality ( like validating the ranges )

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/ReadVectoredTest.java
```diff
@@ -41,6 +60,139 @@ void testVectoredReads(
...
+
+      S3SeekableInputStream s3SeekableInputStream =
+          s3AALClientStreamReader.createReadStream(S3Object.RANDOM_1GB);
+
+      List<ObjectRange> objectRanges = new ArrayList<>();
+      objectRanges.add(new ObjectRange(new CompletableFuture<>(), 700, 500));
+      objectRanges.add(null);
+
+      assertThrows(
+          NullPointerException.class,
```
SanjayMarreddi, line 121: I'm not fully aware of the high-level callers of this `readVectored` API, but are we sure that we are aligned on the higher level exception handling? Like, is `NullPointerException` an expected one or they expected a different one?

For example in `parquet-common` the contract [does not mention](https://www.javadoc.io/doc/org.apache.parquet/parquet-common/latest/org/apache/parquet/io/SeekableInputStream.html#readVectored-java.util.List-org.apache.parquet.bytes.ByteBufferAllocator-) `NullPointerException`

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/ReadVectoredTest.java
```diff
@@ -41,6 +60,139 @@ void testVectoredReads(
...
+      IntFunction<ByteBuffer> allocate = ByteBuffer::allocate;
+
+      S3SeekableInputStream s3SeekableInputStream =
+          s3AALClientStreamReader.createReadStream(S3Object.RANDOM_1GB);
+
+      List<ObjectRange> objectRanges = new ArrayList<>();
+      objectRanges.add(new ObjectRange(new CompletableFuture<>(), 700, 500));
+      objectRanges.add(null);
+
+      assertThrows(
```
fuatbasik, line 120: @ahmarsuhail  are we OK stream throwing NullPointerException? I am not sure if current stream throws that at all or just IOException or EOFException

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/ReadVectoredTest.java
```diff
@@ -15,16 +15,35 @@
...
+import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import software.amazon.s3.analyticsaccelerator.S3SeekableInputStream;
+import software.amazon.s3.analyticsaccelerator.common.ObjectRange;
 
 public class ReadVectoredTest extends IntegrationTestBase {
```
fuatbasik, line 40: It might be good to consider certain test cases that will test known underlying interactions here. For example, test cases where ranges will span multiple blocks or where multiple ranges are accessing to same block.

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/ReadVectoredTest.java
ahmarsuhail: done, thanks

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManager.java
```diff
@@ -229,6 +225,27 @@ public synchronized void makeRangeAvailable(long pos, long len, ReadMode readMod
...
+    if (readMode == ReadMode.READ_VECTORED
+        || readMode == ReadMode.REMAINING_COLUMN_PREFETCH
+        || readMode == ReadMode.COLUMN_PREFETCH
+        || readMode == ReadMode.DICTIONARY_PREFETCH
+        || readMode == ReadMode.PREFETCH_TAIL) {
+      return false;
+    }
+
+    return true;
+  }
```
SanjayMarreddi, line 247: May be a nit, but I think this logic can be simplified if we update our `ReadMode` enum class to something like below: ( The boolean values assigned may be wrong, but syntax should be like that ). Then in this blockManager, we could just use `readMode.allowsSequentialPrefetch()` value and remove the above method.  In the future, if we have more readModes, we dont have to make changes in both enum class and this method.

```
public enum ReadMode {
  SYNC(true),
  ASYNC(true),
  SMALL_OBJECT_PREFETCH(true),
  SEQUENTIAL_FILE_PREFETCH(true),
  READ_VECTORED(false),
  REMAINING_COLUMN_PREFETCH(false),
  COLUMN_PREFETCH(false),
  DICTIONARY_PREFETCH(false),
  PREFETCH_TAIL(false),

  private final boolean allowsSequentialPrefetch;

  ReadMode(boolean allowsSequentialPrefetch) {
    this.allowsSequentialPrefetch = allowsSequentialPrefetch;
  }

  public boolean allowsSequentialPrefetch() {
    return allowsSequentialPrefetch;
  }
}
```

ahmarsuhail: thanks this is a good suggestion :) will do

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/impl/PhysicalIOImpl.java
```diff
@@ -289,6 +297,23 @@ private void readIntoDirectBuffer(ByteBuffer buffer, Blob blob, ObjectRange rang
     }
   }
 
+  /**
+   * Does the block creation for the read vectored ranges.
+   *
+   * @param objectRanges Vectored ranges to fetch
+   */
+  private void makeReadVectoredRangesAvailable(List<ObjectRange> objectRanges) {
```
fuatbasik, line 305: I am concerned about concurrency here. The scenario I have in mind 2 streams against the same S3 key calling readvectored concurrently with overlapping ranges. Would that be a problem?

fuatbasik (commented):
thanks a lot @ahmarsuhail . I put 2-3 comments. I have a question about concurrency in one of my comments if you think that is OK my concern is not valid, i am happy to approve

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/logical/parquet/ParquetPrefetchTailTask.java
```diff
@@ -76,7 +77,7 @@ public List<Range> prefetchTail() {
                 ParquetUtils.getFileTailPrefetchRanges(logicalIOConfiguration, 0, contentLength);
             IOPlan ioPlan = new IOPlan(ranges);
             // Create a non-empty IOPlan only if we have a valid range to work with
-            physicalIO.execute(ioPlan);
+            physicalIO.execute(ioPlan, ReadMode.PREFETCH_TAIL);
```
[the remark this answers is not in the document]
ahmarsuhail: no, single IoPlan should not have multiple readModes. so this is ok here

PR #292, opened 2025-06-17
Move Block constructor to Block Builder

fuatbasik:
## Description of change
Hey, with this change, i am proposing to move away from constructor for Block class and introduced a custom builder. This is because constructor was getting out of help with 8+ parameters, making it hard to navigate with Default values + ordering. 

This will help me to introduce more parameters when adding retry strategies. 
#### Relevant issues
None. 
No. It only touches private surface. 

No.

Updating existing tests to use the new Builder and confirmed all of them are passing. 

- [NA ] I have updated the CHANGELOG or README if appropriate

---

### DIFF
--- a/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
+++ b/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
@@ -28,6 +28,7 @@ import software.amazon.s3.analyticsaccelerator.common.Metrics;
 import software.amazon.s3.analyticsaccelerator.common.Preconditions;
 import software.amazon.s3.analyticsaccelerator.common.telemetry.Operation;
 import software.amazon.s3.analyticsaccelerator.common.telemetry.Telemetry;
+import software.amazon.s3.analyticsaccelerator.io.physical.PhysicalIOConfiguration;
 import software.amazon.s3.analyticsaccelerator.request.GetRequest;
 import software.amazon.s3.analyticsaccelerator.request.ObjectClient;
 import software.amazon.s3.analyticsaccelerator.request.ObjectContent;
@@ -48,10 +49,10 @@ public class Block implements Closeable {
   private final OpenStreamInformation openStreamInformation;
   private final ReadMode readMode;
   private final Referrer referrer;
-  private final long readTimeout;
-  private final int readRetryCount;
   @Getter private final long generation;
-
+  @Getter private final PhysicalIOConfiguration configuration;
+  private final int readRetryCount;
+  private final long readTimeout;
   private final Metrics aggregatingMetrics;
   private final BlobStoreIndexCache indexCache;
   private static final String OPERATION_BLOCK_GET_ASYNC = "block.get.async";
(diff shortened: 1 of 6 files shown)


input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
```diff
@@ -296,10 +291,151 @@ private byte[] getData() throws IOException {
...
+      return this;
+    }
+
+    /**
+     * Builds a new Block instance with the configured parameters.
+     *
+     * @return a new Block instance
+     * @throws IOException if an I/O error occurs during Block creation
+     */
+    public Block build() throws IOException {
```
rajdchak, line 428: Shouldn't we run a validation to check whether all the mandatory fields have been set or not?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
```diff
@@ -296,10 +291,151 @@ private byte[] getData() throws IOException {
...
+
   /** Closes the {@link Block} and frees up all resources it holds */
   @Override
   public void close() {
     // Only the source needs to be canceled, the continuation will cancel on its own
     this.source.cancel(false);
   }
+
+  /** Custom builder for Block class. */
+  public static class BlockBuilder {
```
stubz151, line 311: why are we making a whole class just for a builder with some defaults? can't you just do that with the builder default annotation?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
rajdchak: ignore this, I missed we have null checks in the constructor.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
ozkoca, line 311: Are there any reason to create an inner class instead of using Lombok's `@Builder `annotation?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
```diff
@@ -296,10 +291,151 @@ private byte[] getData() throws IOException {
...
+    private BlockKey blockKey;
+    private ObjectClient objectClient;
+    private Telemetry telemetry = Telemetry.DEFAULT;
+    private long generation = 0;
+    private ReadMode readMode;
+    private PhysicalIOConfiguration configuration = PhysicalIOConfiguration.DEFAULT;
+    private Metrics aggregatingMetrics;
+    private BlobStoreIndexCache indexCache;
+    private OpenStreamInformation openStreamInformation = OpenStreamInformation.DEFAULT;
+
```
rajdchak, line 321: I think there is a risk in setting these to DEFAULT. For eg : what if user has passed a value for openStreamInformation and somehow we missed setting that while creating the block builder, since we are setting these to DEFAULT we will never catch this in the non null check in the constructor and the value set will be not what the user wanted.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
fuatbasik: it is being done at the `private` constructor. I am happy to move them here though if you think it is important..

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
```diff
@@ -94,10 +93,6 @@ public Block(
     Preconditions.checkArgument(0 <= end, "`end` must be non-negative; was: %s", end);
     Preconditions.checkArgument(
         start <= end, "`start` must be less than `end`; %s is not less than %s", start, end);
-    Preconditions.checkArgument(
-        0 < readTimeout, "`readTimeout` must be greater than 0; was %s", readTimeout);
-    Preconditions.checkArgument(
-        0 < readRetryCount, "`readRetryCount` must be greater than 0; was %s", readRetryCount);
```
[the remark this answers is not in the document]
fuatbasik: we are now getting them from configuration and configuration is making sure these values are set properly. We also do not accept these values in the builder, and set them to final to avoid re-setting after object creation.

PR #293, opened 2025-06-20
Add support for readFully at the S3SeekableInputStream level

SanjayMarreddi:
## Description of change
We want to support `readFully` as a part of our ongoing effort to integrate with Iceberg S3FileIO by default. [RangeReadable in Iceberg](https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/io/RangeReadable.java#L47)

**Note:** Here's the comparison with existing read method:

| Aspect | `read(byte[] buffer, int offset, int length)` | `readFully(long position, byte[] buffer, int offset, int length)` |
|--------|----------------------------------------------|-------------------------------------------------------------------|
| Method Signature | Reads from current stream position | Reads from specified position |
| Position Behavior | Advances the stream position by the number of bytes actually read | Does not modify the stream position (position-independent read) |
| Return & Error Handling | Returns `int` indicating bytes read; returns -1 at end of stream; may return fewer bytes than requested | Returns `void`; throws `IOException` if unable to read exact number of requested bytes |

#### Relevant issues
Once this feature is released as `1.2.0`, we will update this [Iceberg PR](https://github.com/apache/iceberg/pull/13361)

No

Yes

Ran existing and new tests locally

---

### DIFF
--- a/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStream.java
+++ b/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStream.java
@@ -15,6 +15,7 @@
  */
 package software.amazon.s3.analyticsaccelerator;
 
+import java.io.EOFException;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.List;
@@ -232,6 +233,41 @@ public class S3SeekableInputStream extends SeekableInputStream {
     Preconditions.checkNotNull(allocate, "allocate must not be null");
 
     logicalIO.readVectored(ranges, allocate);
+  }
+
+  /**
+   * Fill the provided buffer with the contents of the input source starting at {@code position} for
+   * the given {@code offset} and {@code length}.
+   *
+   * @param position start position of the read
+   * @param buffer target buffer to copy data
+   * @param offset offset in the buffer to copy the data
+   * @param length size of the read
+   * @throws IOException if an I/O error occurs
+   */
+  public void readFully(long position, byte[] buffer, int offset, int length) throws IOException {
+    throwIfClosed("cannot read from closed stream");
+    validatePositionedReadArgs(position, buffer, offset, length);
+
+    if (length == 0) {
+      return;
+    }
+
+    this.telemetry.measureVerbose(
+        () ->
+            Operation.builder()
+                .name(OPERATION_READ)
+                .attribute(StreamAttributes.uri(this.s3URI))
+                .attribute(StreamAttributes.etag(this.logicalIO.metadata().getEtag()))
+                .attribute(StreamAttributes.range(position, position + length - 1))
+                .build(),
+        () -> {
+          int bytesRead = this.logicalIO.read(buffer, offset, length, position);
+          if (bytesRead < length) {
+            throw new EOFException(
+                "Reached the end of stream with " + (length - bytesRead) + " bytes left to read");
+          }
+        });
   }
 
   /**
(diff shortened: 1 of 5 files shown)


input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/RandomAccessReadable.java
```diff
@@ -76,4 +76,16 @@ public interface RandomAccessReadable extends Closeable {
...
+   * Fill the provided buffer with the contents of the input source starting at {@code position} for
+   * the given {@code offset} and {@code length}.
+   *
+   * @param position start position of the read
+   * @param buffer target buffer to copy data
+   * @param offset offset in the buffer to copy the data
+   * @param length size of the read
+   * @throws IOException if an I/O error occurs
+   */
+  void readFully(long position, byte[] buffer, int offset, int length) throws IOException;
```
rajdchak, line 90: May be I am missing some knowledge here, what is the difference between the read and readFully apis? What does readFully do differently than read?

SanjayMarreddi: Yeah, I understand that its not obvious.. I have added the differences above and also linked the interface from Iceberg.

stubz151 (approved):

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStream.java
```diff
@@ -234,6 +235,41 @@ public void readVectored(
...
+   * @param buffer target buffer to copy data
+   * @param offset offset in the buffer to copy the data
+   * @param length size of the read
+   * @throws IOException if an I/O error occurs
+   */
+  public void readFully(long position, byte[] buffer, int offset, int length) throws IOException {
+    throwIfClosed("cannot read from closed stream");
+    validatePositionedReadArgs(position, buffer, offset, length);
+
+    if (length == 0) {
```
rajdchak, line 252: should we check this as a precondition that length should be > 0?

rajdchak, line 252: should we check this as a precondition that length should be > 0?

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStreamTest.java
```diff
@@ -448,6 +448,84 @@ public void testInsufficientBuffer() throws IOException {
...
+      assertThrows(IllegalArgumentException.class, () -> stream.readFully(-1, new byte[10], 0, 5));
+      // -1 is invalid length
+      assertThrows(IllegalArgumentException.class, () -> stream.readFully(0, new byte[10], 0, -1));
+      // Requesting more data than byte buffer size
+      assertThrows(IndexOutOfBoundsException.class, () -> stream.readFully(0, new byte[5], 0, 10));
+    }
+  }
+
+  @Test
+  void testReadFullyHappyCase() throws IOException {
```
rajdchak, line 470: can we also add a test case for the case when exception is thrown when bytesToRead < length

SanjayMarreddi: Good point, added it now!

rajdchak (approved):
Thank @SanjayMarreddi looks good to me!

PR #298, opened 2025-06-25
IoStats-CallBack-AAL

vaibhav5140:
## Description of change

This PR implements a callback mechanism that allows the analytics accelerator to pass metrics back to the S3A connector, ensuring that each GET operation properly increments the STREAM_READ_ANALYTICS_GET_REQUESTS counter. This enables S3A to monitor and assert with accurate statistics. 

Public Jira: https://issues.apache.org/jira/browse/HADOOP-19364

Haddop S3A PR: https://github.com/apache/hadoop/pull/7763

---

### DIFF
--- a/common/src/main/java/software/amazon/s3/analyticsaccelerator/util/DefaultRequestCallbackImpl.java
+++ b/common/src/main/java/software/amazon/s3/analyticsaccelerator/util/DefaultRequestCallbackImpl.java
@@ -24,11 +24,11 @@ public class DefaultRequestCallbackImpl implements RequestCallback {
 
   @Override
   public void onGetRequest() {
-    LOG.debug("GET request made");
+    LOG.trace("GET request made");
   }
 
   @Override
   public void onHeadRequest() {
-    LOG.debug("HEAD request made");
+    LOG.trace("HEAD request made");
   }
 }
--- a/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStreamFactory.java
+++ b/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStreamFactory.java
@@ -80,7 +80,8 @@ public class S3SeekableInputStreamFactory implements AutoCloseable {
     this.parquetColumnPrefetchStore =
         new ParquetColumnPrefetchStore(configuration.getLogicalIOConfiguration());
     this.objectMetadataStore =
-        new MetadataStore(objectClient, telemetry, configuration.getPhysicalIOConfiguration());
+        new MetadataStore(
+            objectClient, telemetry, configuration.getPhysicalIOConfiguration(), metrics);
     this.objectFormatSelector = new ObjectFormatSelector(configuration.getLogicalIOConfiguration());
     this.objectBlobStore =
         new BlobStore(objectClient, telemetry, configuration.getPhysicalIOConfiguration(), metrics);
--- a/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/MetadataStore.java
+++ b/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/MetadataStore.java
@@ -25,12 +25,14 @@ import java.util.concurrent.CompletableFuture;
 import lombok.NonNull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import software.amazon.s3.analyticsaccelerator.common.Metrics;
 import software.amazon.s3.analyticsaccelerator.common.telemetry.Operation;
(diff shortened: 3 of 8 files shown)


input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStream.java
```diff
@@ -51,19 +53,25 @@ public class S3SeekableInputStream extends SeekableInputStream {
...
    * @param logicalIO already initialised LogicalIO
    * @param telemetry The {@link Telemetry} to use to report measurements.
+   * @param openStreamInformation The to use to report measurements.
    */
   S3SeekableInputStream(
-      @NonNull S3URI s3URI, @NonNull LogicalIO logicalIO, @NonNull Telemetry telemetry) {
+      @NonNull S3URI s3URI,
+      @NonNull LogicalIO logicalIO,
+      @NonNull Telemetry telemetry,
+      @NonNull OpenStreamInformation openStreamInformation) {
```
rajdchak, line 70: I don't think passing of openStreamInformation is required here. Not sure why are we doing this?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStreamFactory.java
```diff
@@ -98,7 +98,8 @@ public S3SeekableInputStreamFactory(
    * @return An instance of the input stream.
```
ahmarsuhail, line 98: also shouldn't need any changes here

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
```diff
@@ -126,7 +126,9 @@ private void generateSourceAndData() throws IOException {
                 .etag(this.blockKey.getObjectKey().getEtag())
                 .referrer(referrer)
                 .build();
-
+        if (openStreamInformation.getRequestCallback() != null) {
```
ahmarsuhail, line 129: you shouldn't need a null check here. the default implementation of this should be a NO-OP. 

S3A will overwrite the callbacks, so this actually does something

vaibhav5140: Addressed in updated PR

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStream.java
vaibhav5140: Addressed in updated revision

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStreamFactory.java
vaibhav5140: Addressed in updated revision

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/MetadataStore.java
```diff
@@ -125,8 +125,13 @@ public synchronized CompletableFuture<ObjectMetadata> asyncGet(
                         .name(OPERATION_METADATA_HEAD_ASYNC)
                         .attribute(StreamAttributes.uri(s3URI))
                         .build(),
-                objectClient.headObject(
-                    HeadRequest.builder().s3Uri(s3URI).build(), openStreamInformation)));
+                () -> {
+                  CompletableFuture<ObjectMetadata> result =
```
ahmarsuhail, line 129: what about updating the head request metric?

common/src/main/java/software/amazon/s3/analyticsaccelerator/util/DefaultRequestCallbackImpl.java
```diff
@@ -0,0 +1,34 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Default implementation of {@link RequestCallback} that logs requests at debug level. */
+public class DefaultRequestCallbackImpl implements RequestCallback {
+  private static final Logger LOG = LoggerFactory.getLogger(DefaultRequestCallbackImpl.class);
+
+  @Override
+  public void onGetRequest() {
+    LOG.debug("GET request made");
```
ahmarsuhail, line 27: sorry can you move this to TRACE? just thinking logging this at debug will create a lot of noise in the benchmarking logs

ahmarsuhail (dismissed):
+1, pending moving logging to trace. 

would be great if you can also update the HEAD request metric also. we can move it into objectClient later

ahmarsuhail (approved):
+1, LGTM

PR #307, opened 2025-07-07
Add Retries and Retry Policy

fuatbasik:
## Description of change
This is the first PR of a series of 2, that is adding RetryPolicy (with Failsafe as a dependency) and applying these retries with the current defaults. In the next PR, I will add the logic to pass RetryPolicy from library user to block, through Open Stream Information 

#### Relevant issues
NO. 

No. 

Yes, it introduces a new capability to change/add retries on top of SDK retries. 

Added new unit tests, confirmed existing GrayFailureTests are passing. 

Yes.  I will update the ChangeLog and Readme in the next PR. 

---


input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
```diff
@@ -106,80 +103,61 @@ public Block(
     this.openStreamInformation = openStreamInformation;
     this.readMode = readMode;
     this.referrer = new Referrer(this.blockKey.getRange().toHttpString(), readMode);
-    this.readTimeout = readTimeout;
-    this.readRetryCount = readRetryCount;
     this.aggregatingMetrics = aggregatingMetrics;
     this.indexCache = indexCache;
-    generateSourceAndData();
+    this.configuration = configuration;
```
stubz151, line 108: is this variable ever used? 
think line 111 and 109-110 use class variable vs constructor variable in different ways maybe just remove this config?

input-stream/build.gradle.kts
```diff
@@ -73,6 +73,7 @@ dependencies {
     implementation(libs.parquet.format)
     implementation(libs.slf4j.api)
     implementation(libs.caffeine)
+    implementation(libs.failsafe)
```
stubz151, line 76: just checking if u ran jar tvf and saw it pull in any other dependencies?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/retry/RetryExecutor.java
```diff
@@ -0,0 +1,85 @@
...
+   *
+   * @param runnable the operation to execute, which may throw IOException
+   * @throws IOException if the operation fails after all retry attempts are exhausted
+   * @throws NullPointerException if runnable is null
+   */
+  void executeWithRetry(IORunnable runnable) throws IOException;
+
+  /**
+   * Executes a supplier with retry logic according to the configured retry policy.
+   *
```
stubz151, line 46: these 2 lines are basically saying the same thing about using the retry policy maybe just merge them?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/retry/RetryExecutor.java
```diff
@@ -0,0 +1,85 @@
...
+   * @param supplier the operation to execute, returns <R>, may throw IOException
+   * @return R result of the supplier.
+   * @throws IOException if the operation fails due to an I/O error
+   */
+  R getWithRetry(IOSupplier<R> supplier) throws IOException;
+
+  /** Functional interface for operations that can throw IOException. */
+  @FunctionalInterface
+  interface IORunnable {
```
stubz151, line 59: think this interface and the interface below can be handy in the future should we move them to their own classes.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/retry/SeekableInputStreamRetryExecutor.java
```diff
@@ -0,0 +1,128 @@
...
+  /**
+   * Creates a retry executor with a single retry policy based on the provided configuration.
+   *
+   * <p>The retry policy is configured using settings from the PhysicalIOConfiguration, including
+   * retry count and timeout values.
+   *
+   * @param configuration the physical IO configuration containing retry settings
+   * @throws NullPointerException if configuration is null
+   */
+  public SeekableInputStreamRetryExecutor(PhysicalIOConfiguration configuration) {
```
stubz151, line 65: I don't think this should take PhysicalIOConfiguration, I can see us wanting to re-use this logic elsewhere and we might not necessarily want to use PhysicalIOConfiguration there. Can't we split this out to it's own thing? even if it just includes the 2 settings we have for now.

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/retry/RetryPolicyBuilderTest.java
```diff
@@ -0,0 +1,105 @@
...
+    RetryPolicyBuilder<String> builder = new RetryPolicyBuilder<>();
+    Duration timeout = Duration.ofSeconds(10);
+    RetryPolicyBuilder<String> result = builder.withMaxDuration(timeout);
+
+    assertSame(builder, result);
+    assertNotNull(builder.build());
+  }
+
+  @Test
+  void testHandleSingleException() {
```
stubz151, line 66: lots of these tests are very similar is there the possibility of us combining them into a parameterized test maybe?

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/retry/SeekableInputStreamRetryExecutorTest.java
```diff
@@ -0,0 +1,205 @@
...
+class SeekableInputStreamRetryExecutorTest {
+
+  @Test
+  void testNoArgsConstructor() {
+    SeekableInputStreamRetryExecutor<String> executor = new SeekableInputStreamRetryExecutor<>();
+    assertNotNull(executor);
+  }
+
+  @Test
+  void testConfigurationConstructor() {
```
stubz151, line 34: same thing about using parameterized test to just cover all the constructor tests.

input-stream/build.gradle.kts
fuatbasik: failsafe itself is dependency free, but i will double check the jar.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
```diff
@@ -110,76 +112,80 @@ public Block(
...
-                () -> {
-                  this.aggregatingMetrics.add(MetricKey.GET_REQUEST_COUNT, 1);
-                  return objectClient.getObject(getRequest, openStreamInformation);
-                });
+  /**
+   * Helper to construct source and data
+   *
+   * @throws RuntimeException if all retries fails and an error occurs
+   */
+  private void generateSourceAndData() {
```
fuatbasik, line 144: With this PR we are changing the retry behaviour, intentionally. Today if generateSourceAndData timeouts (mostly this is relevant to StreamUtils.toByteArray()), we will retry, even if there is no call to 'read' function of that block. With this change we will only retry if there is a call to 'read' function. This could be when we prefetch a block but never read it, but it does make sense not to retry them anyways.

stubz151 (approved):

stubz151:
LGTM TY

PR #319, opened 2025-07-18
Upgrade SDK version and fix user-agent changes to avoid prepending empty string

fuatbasik:
This PR does 2 things 
1/It updates SDK version to the same version as Iceberg. This is because SDK introduced a breaking change on user-agent evaluation. 
2/It adds a small check to User agent, does not prepend if string is empty. It confirms client can still make requests with different ObjectClientConfigurations (minimal checks to be fair) to ensure we can still read objects with different configurations.
3/ It adds CRCCheck class, taken out from SDK to confirm streams read from S3 are identical with and without AAL. 

---

### DIFF
--- a/object-client/src/main/java/software/amazon/s3/analyticsaccelerator/request/UserAgent.java
+++ b/object-client/src/main/java/software/amazon/s3/analyticsaccelerator/request/UserAgent.java
@@ -39,11 +39,13 @@ public final class UserAgent {
    * @param userAgentPrefix to prepend the default user-agent string
    */
   public void prepend(String userAgentPrefix) {
-    if (Objects.nonNull(userAgentPrefix) && !userAgentPrefix.isEmpty())
+    if (Objects.nonNull(userAgentPrefix) && !userAgentPrefix.trim().isEmpty()) {
       this.userAgent = sanitizeInput(userAgentPrefix) + " " + this.userAgent;
+    }
   }
 
   private static String sanitizeInput(String input) {
-    return input == null ? "unknown" : input.replaceAll(UA_DENYLIST_REGEX, "_");
+    if (Objects.isNull(input)) return "unknown";
+    return input.trim().replaceAll(UA_DENYLIST_REGEX, "_");
   }
 }
--- a/input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/SDKClientTest.java
+++ b/input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/SDKClientTest.java
@@ -18,7 +18,11 @@ package software.amazon.s3.analyticsaccelerator.access;
 import static org.junit.jupiter.api.Assertions.*;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Stream;
 import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
 import software.amazon.s3.analyticsaccelerator.ObjectClientConfiguration;
 import software.amazon.s3.analyticsaccelerator.S3SdkObjectClient;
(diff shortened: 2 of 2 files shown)


input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/ChecksumAssertions.java
```diff
@@ -17,16 +17,14 @@
...
-
 public class ChecksumAssertions {
   /**
    * Checksum assertion
    *
    * @param expected expected checksum
    * @param actual actual checksum
    */
-  public static void assertChecksums(SdkChecksum expected, SdkChecksum actual) {
+  public static void assertChecksums(Crc32CChecksum expected, Crc32CChecksum actual) {
```
rajdchak, line 27: sorry missing some knowledge here, why do we need to change this to Crc32CChecksum?

fuatbasik: thanks for asking. SDK deprecated Checksum algorithms, therefore, i took the same algorithm out of SDK and added to our code-base.

object-client/src/main/java/software/amazon/s3/analyticsaccelerator/request/UserAgent.java
```diff
@@ -39,7 +39,7 @@ public final class UserAgent {
    * @param userAgentPrefix to prepend the default user-agent string
    */
   public void prepend(String userAgentPrefix) {
-    if (Objects.nonNull(userAgentPrefix))
+    if (Objects.nonNull(userAgentPrefix) && !userAgentPrefix.isEmpty())
```
vaibhav5140, line 42: I think we can just use StringUtils.isNotEmpty(userAgentPrefix) instead of this

vaibhav5140: Can we use StringUtils.isNotBlank(userAgentPrefix) as this will handle whitespace only strings (like "   ", "\t") as well?

fuatbasik: StringUtils is an Apache Commons dependency but checking for whitespace characters is a really big call-out. Thanks.

object-client/src/main/java/software/amazon/s3/analyticsaccelerator/request/UserAgent.java
```diff
@@ -39,8 +39,10 @@ public final class UserAgent {
    * @param userAgentPrefix to prepend the default user-agent string
    */
   public void prepend(String userAgentPrefix) {
-    if (Objects.nonNull(userAgentPrefix))
+    if (Objects.nonNull(userAgentPrefix) && !userAgentPrefix.trim().isEmpty()) {
+      userAgentPrefix = userAgentPrefix.trim();
       this.userAgent = sanitizeInput(userAgentPrefix) + " " + this.userAgent;
```
stubz151, line 44: we have a sanitizeInput() method shouldn't that just trim it ?

vaibhav5140: +1

object-client/src/main/java/software/amazon/s3/analyticsaccelerator/request/UserAgent.java
```diff
@@ -39,11 +39,14 @@ public final class UserAgent {
...
+    if (Objects.nonNull(userAgentPrefix) && !userAgentPrefix.trim().isEmpty()) {
       this.userAgent = sanitizeInput(userAgentPrefix) + " " + this.userAgent;
+    }
   }
 
   private static String sanitizeInput(String input) {
-    return input == null ? "unknown" : input.replaceAll(UA_DENYLIST_REGEX, "_");
+    if (Objects.isNull(input)) return "unknown";
+    input = input.trim();
+    return input.replaceAll(UA_DENYLIST_REGEX, "_");
```
stubz151, line 50: does input = input.trim().replaceAll(UA_DENYLIST_REGEX, "_");

not work?

PR #320, opened 2025-07-23
Introducing retry policy to new PhysicalIO

ozkoca:
## Description of change
This PR adds the capability of retry to the new PhysicalIO 

#### Relevant issues
https://github.com/awslabs/analytics-accelerator-s3/pull/286
https://github.com/awslabs/analytics-accelerator-s3/pull/287
https://github.com/awslabs/analytics-accelerator-s3/pull/288
https://github.com/awslabs/analytics-accelerator-s3/pull/289
https://github.com/awslabs/analytics-accelerator-s3/pull/294
https://github.com/awslabs/analytics-accelerator-s3/pull/316

#### Does this contribution introduce any breaking changes to the existing APIs or behaviors?
No

#### Does this contribution introduce any new public APIs or behaviors?
No

#### How was the contribution tested?
Unit tests

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and I agree to the terms of the [Developer Certificate of Origin (DCO)](https://developercertificate.org/).

PR #287, opened 2025-06-09
[New Physical IO] Adopting OpenStremInformation changes to new PhysicalIO

ozkoca:
## Description of change
- This change adopts the changes from [PR](https://github.com/awslabs/analytics-accelerator-s3/pull/283) to the new Physical IO implementation.
- Updates comment in DataBlock object

#### Relevant issues
[OpenStremInformation PR](https://github.com/awslabs/analytics-accelerator-s3/pull/283) 
[Initial version of Physical IO](https://github.com/awslabs/analytics-accelerator-s3/pull/286)

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and I agree to the terms of the [Developer Certificate of Origin (DCO)](https://developercertificate.org/).

PR #286, opened 2025-06-06
[INITIAL VERSION] New PhysicalIO

ozkoca:
## Description of change
This PR includes the initial changes required for new PhysicalIO. The new objects are representative of the existing objects like DataBlock is the new version of Block object which will be renamed at the end of the integration. 

Next Steps:
- Add unit tests for the new objects
- Update Javadoc
- Add enhanced features read ahead, sequential prefetching, range splitting, small object preetching, etc.
- Adopt memory manager
- Exception handling, retries and timeouts.
- Integrating new PhysicalIO with Blob object

#### Does this contribution introduce any breaking changes to the existing APIs or behaviors?
No

#### Does this contribution introduce any new public APIs or behaviors?
No

#### How was the contribution tested?
Integrated with existing system in a separate branch and ran tests

#### Does this contribution need a changelog entry?
N/A

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and I agree to the terms of the [Developer Certificate of Origin (DCO)](https://developercertificate.org/).

PR #325, opened 2025-07-25
New PhysicalIO implementation

ozkoca:
## Description of change
This PR implements a new PhysicalIO design with key improvements:

**Fixed Block Size:** Previously, block sizes varied based on request ranges, requiring entire request ranges to complete before blocks became available. The new design uses fixed-size blocks that become ready as soon as individual blocks are filled, enabling faster data access and better parallelization.

**Direct Block Writing:** Eliminates an extra memory copy by writing S3 data directly into Block storage instead of copying from intermediate buffers, reducing memory overhead and CPU usage.

**Improved Concurrency:** Fixed-size blocks allow multiple blocks to be processed independently, improving throughput for concurrent read operations.

**Better Memory Management:** Predictable block sizes enable more efficient memory allocation and cache management strategies.

**Enhanced Read Performance:** Blocks become available for reading as soon as they're filled, rather than waiting for entire request ranges to complete, reducing read latency.

#### Relevant issues
https://github.com/awslabs/analytics-accelerator-s3/pull/286
https://github.com/awslabs/analytics-accelerator-s3/pull/287
https://github.com/awslabs/analytics-accelerator-s3/pull/288
https://github.com/awslabs/analytics-accelerator-s3/pull/289
https://github.com/awslabs/analytics-accelerator-s3/pull/294
https://github.com/awslabs/analytics-accelerator-s3/pull/316
https://github.com/awslabs/analytics-accelerator-s3/pull/320
https://github.com/awslabs/analytics-accelerator-s3/pull/323
https://github.com/awslabs/analytics-accelerator-s3/pull/324

No

No

Unit tests, microbenchmarks

---

### DIFF
--- a/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
+++ b/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
@@ -268,7 +268,7 @@ public class PhysicalIOConfiguration {
     Preconditions.checkArgument(readBufferSize > 0, "`readBufferSize` must be positive");
     Preconditions.checkArgument(targetRequestSize > 0, "`targetRequestSize` must be positive");
     Preconditions.checkArgument(
-        requestToleranceRatio >= 1, "`requestToleranceRatio` must be greater or equal than 1");
+        requestToleranceRatio >= 1, "`requestToleranceRatio` must be greater than or equal than 1");
 
     this.memoryCapacityBytes = memoryCapacityBytes;
     this.memoryCleanupFrequencyMilliseconds = memoryCleanupFrequencyMilliseconds;
--- a/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/reader/StreamReader.java
+++ b/input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/reader/StreamReader.java
@@ -410,7 +410,10 @@ public class StreamReader implements Closeable {
         blocks.stream().filter(block -> !block.isDataReady()).collect(Collectors.toList()));
   }
 
-  /** Shuts down the StreamReader object. */
+  /**
+   * Releases any resources held by this StreamReader. Currently, no cleanup is required, but this
+   * method is provided for future extensibility.
+   */
   @Override
   public void close() {}
 }


stubz151:
I know there are multiple sub pr's but can you update the description for this one with some details on the overall changes and decisions etc. Am struggling to review this and that might help me.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
```diff
@@ -228,65 +174,105 @@ public int read(byte @NonNull [] buf, int off, int len, long pos) throws IOExcep
...
+
+  /**
+   * Waits for the block's data to become available. This method blocks until {@link
+   * #setData(byte[])} is called.
+   *
+   * @throws IOException if the thread is interrupted or data is not set
+   */
+  private void awaitData() throws IOException {
+    try {
+      if (!dataReadyLatch.await(readTimeout, TimeUnit.MILLISECONDS)) {
```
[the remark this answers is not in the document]
ozkoca: If we don't set the timeout for await, it will wait forever to countdownlatch to be zero. This timeout prevents this happening

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Blob.java
```diff
@@ -94,7 +95,15 @@ public int read(long pos) throws IOException {
     try {
       lock.readLock().lock();
       blockManager.makePositionAvailable(pos, ReadMode.SYNC);
-      return blockManager.getBlock(pos).get().read(pos);
+      Optional<Block> block = blockManager.getBlock(pos);
+      if (!block.isPresent()) {
+        throw new IllegalStateException(
```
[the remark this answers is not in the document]
ozkoca: Normally we shouldn't face this error and I kept it as it is with the existing system

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManager.java
```diff
@@ -240,9 +265,76 @@ private long truncatePos(long pos) {
...
+  }
+
+  /**
+   * Retrieves the {@link Block} containing the given position, if it exists in the block store.
+   *
+   * @param pos the byte position within the object to look up
+   * @return an {@link Optional} containing the {@link Block} if present; otherwise, {@link
+   *     Optional#empty()}
+   */
+  public synchronized Optional<Block> getBlock(long pos) {
```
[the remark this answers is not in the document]
ozkoca: if it is not synchronised, BlockManager object may return blocks which are being removed by removeBlocks method

ahmarsuhail (commented):
Thanks a lot @ozkoca, really like these changes! 

I don't understand the retry and block removal stuff completely. For example, what happens when you're reading blocks [1, 2, 3, 4] from a single GET, and the read fails on block 3. I think you'll remove [3, 4], but what does this mean for the calling application? Will discuss with you. 

Also concerned about what these changes mean for sequential prefetching. 

I have not reviewed test code yet, just trying understand the actual code for now. Java docs with more examples will be helpful.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
```diff
@@ -47,6 +45,9 @@ public class PhysicalIOConfiguration {
   private static final boolean DEFAULT_SMALL_OBJECTS_PREFETCHING_ENABLED = true;
   private static final long DEFAULT_SMALL_OBJECT_SIZE_THRESHOLD = 8 * ONE_MB;
   private static final int DEFAULT_THREAD_POOL_SIZE = 96;
+  private static final long DEFAULT_READ_BUFFER_SIZE = 128 * ONE_KB;
+  private static final long DEFAULT_TARGET_REQUEST_SIZE = 8 * ONE_MB;
+  private static final double DEFAULT_REQUEST_TOLERANCE_RATIO = 1.4;
```
vaibhav5140, line 50: Why are we changing the default read buffer size to 128mb? and why request tolerance to 1.4?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManager.java
```diff
@@ -240,9 +265,76 @@ private long truncatePos(long pos) {
...
+  /** cleans data from memory */
+  public void cleanUp() {
+    this.blockStore.cleanUp();
+  }
+
   /** Closes the {@link BlockManager} and frees up all resources it holds */
   @Override
   public void close() {
     blockStore.close();
+    streamReader.close();
```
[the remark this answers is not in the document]
ozkoca: It is not doing anything for now. I just wanted to keep it for the future use

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/SequentialFileTypeIntegrationTest.java
```diff
@@ -55,23 +55,20 @@ void testingSequentialReadPatterBelowPartitionSize(S3ClientKind clientKind, S3Ob
...
+          2,
           s3AALClientStreamReader
               .getS3SeekableInputStreamFactory()
               .getMetrics()
               .get(MetricKey.GET_REQUEST_COUNT));
-
-      // TODO: This should be fixed with the new PhysicalIO, currently the cache hit metric is
-      // inaccurate.
-      //  assertEquals(expectedGETCount,
-      // s3AALClientStreamReader.getS3SeekableInputStreamFactory().getMetrics().get(MetricKey.CACHE_HIT));
```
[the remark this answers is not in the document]
ozkoca: due to performance issues we faced we are not publishing CACHE_HIT and CACHE_MISS metrics in new PhysicalIO

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/RangeOptimiser.java
```diff
@@ -15,54 +15,187 @@
...
+  }
 
-    return generatedRanges;
+  private void mergeSmallFinalChunk(List<List<Integer>> chunks, int maxBlocksBeforeSplit) {
+    if (chunks.size() < 2) return;
+
+    List<Integer> lastChunk = chunks.get(chunks.size() - 1);
+    List<Integer> previousChunk = chunks.get(chunks.size() - 2);
+
+    if (lastChunk.size() < previousChunk.size()
```
vaibhav5140, line 195: Do we need this condition `lastChunk.size() < previousChunk.size()`?, it seems we could merge chunks regardless of their relative sizes, as long as the combined size is within `maxBlocksBeforeSplit`

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
vaibhav5140: Was this tolerance number of 1.4 is based on some POC's?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManager.java
```diff
@@ -101,137 +118,145 @@ private void prefetchSmallObject() {
...
+   * @param pos position of the read
+   * @param readMode whether this ask corresponds to a sync or async read
+   * @return generation of the block
+   */
+  private long getGeneration(long pos, ReadMode readMode) {
+    // Generation is zero for read modes which not allow request extension or first block of the
+    // object
+    if (!readMode.allowRequestExtension() || pos < configuration.getReadBufferSize()) return 0;
+
+    Optional<Block> previousBlock = blockStore.getBlock(pos - 1);
```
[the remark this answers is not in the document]
ozkoca: You are right. We may need to tune `sequentialprefetch.base` or `sequentialprefetch.speed` configurations or even change the logic in the method according to performance results. Noted down for future tasks

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
```diff
@@ -228,65 +169,106 @@ public int read(byte @NonNull [] buf, int off, int len, long pos) throws IOExcep
...
+   * Sets the data for this block and signals that the data is ready for reading. This method should
+   * be called exactly once per block.
    *
-   * @return the bytes fetched by the issued {@link GetRequest}.
-   * @throws IOException if an I/O error occurs
+   * @param data the byte array representing the block's data
+   */
+  public void setData(final byte[] data) {
+    this.data = data;
+    this.aggregatingMetrics.add(MetricKey.MEMORY_USAGE, data.length);
```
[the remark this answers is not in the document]
ozkoca: Both approaches are valid, each with different trade-offs:

**Increasing metric during block creation (early reservation):**
- Reserves memory capacity upfront before data is actually loaded
- May lead to over-reservation since we count memory usage before it's actually consumed
- Prevents memory limit breaches but could cause unnecessary capacity constraints

**Increasing metric in setData (actual usage tracking):**
- Accurately reflects real memory consumption only after data is loaded
- May temporarily exceed memory limits during the gap between data loading and metric updates
- Provides more precise memory accounting but risks brief capacity overruns

The current implementation uses the second approach (setData) to ensure metrics reflect actual memory usage rather than projected usage, accepting the small risk of temporary limit exceedance for better accuracy.

We can discuss both approaches or a different one if you think as an alternative

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/reader/StreamReader.java
```diff
@@ -0,0 +1,416 @@
...
+   */
+  private void removeNonFilledBlocksFromStore(List<Block> blocks) {
+    // Filter out blocks that don't have data and remove them via callback
+    this.removeBlocksFunc.accept(
+        blocks.stream().filter(block -> !block.isDataReady()).collect(Collectors.toList()));
+  }
+
+  /** Shuts down the StreamReader object. */
+  @Override
+  public void close() {}
```
vaibhav5140, line 415: Do we need this method? If it's for future use, please add comment

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockStore.java
```diff
@@ -16,137 +16,147 @@
...
-   * @param pos a byte position
-   * @return the position of the next available byte or empty if there is no next available byte
+   * @param block the {@code Block} to add
    */
-  public OptionalLong findNextLoadedByte(long pos) {
-    Preconditions.checkArgument(0 <= pos, "`pos` must not be negative");
-
-    if (getBlock(pos).isPresent()) {
-      return OptionalLong.of(pos);
+  public void add(Block block) {
```
vaibhav5140, line 114: It might be worth considering what to do if a block with a different range but the same index is added

ozkoca: This is not possible because same index can always store the same range. Do you have a case for your comment?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManager.java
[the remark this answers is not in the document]
fuatbasik: today this is sync. already but i believe we can avoid this going forward. I assume this would need additional changes, therefore, shall we try to address this as a follow-up PR

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
```diff
@@ -17,103 +17,97 @@
...
+   * The underlying byte array containing this block's data. It is set asynchronously via {@link
+   * #setData(byte[])} and should only be accessed through read methods after {@link #awaitData()}
+   * confirms readiness.
+   *
+   * <p>This field is marked {@code @Nullable} because the data is not initialized at construction
+   * time, which would otherwise cause static code analysis to fail.
+   */
+  @Nullable private byte[] data;
+
+  @Nullable private IOException error;
```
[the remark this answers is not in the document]
fuatbasik: Assume there are 2 threads waiting at awaitData. An error occurred and StreamReader set the exception, decrement latch. Now both threads should check the IOException, one can see the exception, looking to Memory other cannot because it is looking at CPU cache.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockStore.java
```diff
@@ -16,137 +16,147 @@
...
+  public void remove(Block block) {
+    if (block == null) {
+      return; // no-op on null input
     }
 
-    return nextMissingByte <= getLastObjectByte()
-        ? OptionalLong.of(nextMissingByte)
-        : OptionalLong.empty();
+    int blockIndex = getBlockIndex(block);
+    if (blocks.remove(blockIndex) != null && block.isDataReady()) {
```
fuatbasik, line 134: would we leak resources here if Blocks data is not ready yet?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/reader/StreamReader.java
```diff
@@ -0,0 +1,419 @@
...
+  }
+
+  /**
+   * Helper to construct retryStrategy
+   *
+   * @return a {@link RetryStrategy} to retry when timeouts are set
+   * @throws RuntimeException if all retries fails and an error occurs
+   */
+  @SuppressWarnings("unchecked")
+  private RetryStrategy createRetryStrategy() {
```
fuatbasik, line 116: we probably need to move this to Utils.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/reader/StreamReader.java
```diff
@@ -0,0 +1,419 @@
...
+                    .attribute(
+                        StreamAttributes.effectiveRange(
+                            blocks.get(0).getBlockKey().getRange().getStart(),
+                            blocks.get(blocks.size() - 1).getBlockKey().getRange().getEnd()))
+                    .build(),
+            () -> {
+              // Calculate the byte range needed to cover all blocks
+              Range requestRange = computeRange(blocks);
+
+              // Build S3 GET request with range, ETag validation, and referrer info
```
fuatbasik, line 184: nit: doesn't have to be S3

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/reader/StreamReader.java
```diff
@@ -0,0 +1,419 @@
...
+                objectContent = this.retryStrategy.get(() -> fetchObjectContent(getRequest));
+              } catch (IOException e) {
+                LOG.error("IOException while fetching object content", e);
+                setErrorOnBlocksAndRemove(blocks, e);
+                return;
+              }
+
+              openStreamInformation.getRequestCallback().onGetRequest();
+
+              if (objectContent == null) {
```
fuatbasik, line 205: in what condition this if will return true? Wouldn't catch clause above handle this?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/reader/StreamReader.java
```diff
@@ -0,0 +1,419 @@
...
+                // Couldn't successfully get the response from S3.
+                // Remove blocks from store and complete async operation
+                removeNonFilledBlocksFromStore(blocks);
+                return;
+              }
+
+              // Process the input stream and populate data blocks
+              try (InputStream inputStream = objectContent.getStream()) {
+                boolean success =
+                    readBlocksFromStream(inputStream, blocks, requestRange.getStart());
```
fuatbasik, line 215: i am a bit confused where the actual reading from network happens. On fetchObjectContent or readBlocksFromStream?

fuatbasik: Whichever is reading bytes from network, should be one we retry

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Blob.java
```diff
@@ -134,7 +144,7 @@ public int read(byte[] buf, int off, int len, long pos) throws IOException {
                         new IllegalStateException(
                             String.format(
                                 "This block object key %s (for position %s) should have been available.",
-                                objectKey.getS3URI().toString(), nextPositionFinal)));
+                                objectKey.getS3URI(), nextPositionFinal)));
```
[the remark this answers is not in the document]
ozkoca: Yes, S£URI has already a toString() [method](https://github.com/awslabs/analytics-accelerator-s3/blob/main/common/src/main/java/software/amazon/s3/analyticsaccelerator/util/S3URI.java#L56).

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/S3SeekableInputStreamFactory.java
```diff
@@ -84,12 +84,17 @@ public S3SeekableInputStreamFactory(
         new MetadataStore(
             objectClient, telemetry, configuration.getPhysicalIOConfiguration(), metrics);
     this.objectFormatSelector = new ObjectFormatSelector(configuration.getLogicalIOConfiguration());
-    this.objectBlobStore =
-        new BlobStore(objectClient, telemetry, configuration.getPhysicalIOConfiguration(), metrics);
     // TODO: calling applications should be able to pass in a thread pool if they so wish
     this.threadPool =
```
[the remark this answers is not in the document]
ozkoca: Will follow-up in a separate PR

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
```diff
@@ -228,65 +166,92 @@ public int read(byte @NonNull [] buf, int off, int len, long pos) throws IOExcep
...
-   * Returns the bytes fetched by the issued {@link GetRequest}. This method will block until the
-   * data is fully available.
+   * Sets the data for this block and signals that the data is ready for reading. This method should
+   * be called exactly once per block.
    *
-   * @return the bytes fetched by the issued {@link GetRequest}.
-   * @throws IOException if an I/O error occurs
+   * @param data the byte array representing the block's data
+   */
+  public void setData(final byte[] data) {
```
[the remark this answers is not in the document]
ozkoca: Added onError flag.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/reader/StreamReader.java
ozkoca: There is no real life use-case for this but added as defensive

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/reader/StreamReader.java
ozkoca: We are creating retryStartegy in 2 places 1/StreamReader and 2/Block. Their handled exceptions are different and in the future we may add different onRetry for these functions. So, we can keep them separated for now.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/Block.java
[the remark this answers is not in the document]
fuatbasik: I think this is on me and i am having second thoughts here. We have 2 places for customers to pass us information about retries now. 1/With Configuration that has ReadTimeOut and ReadRetryCount. And 2/ with RetryStrategy. So far I was thinking we should merge two but looking more into usage patters maybe 2 should overwrite 1 if provided. WDYT?

fuatbasik (approved):

ahmarsuhail (approved):

PR #338, opened 2025-07-31
[feat]:add ttl to metadata

vaibhav5140:
## Description of change
- Add TTL support to MetadataStore using Caffeine cache
- Replace LinkedHashMap with Caffeine cache in MetadataStore
- Add configurable TTL (default: 24 hours) for metadata entries
- Maintain existing size limit (5000 entries) as safety net

Will add a follow up PR to add documentation and explaining edge cases

---


stubz151 (changes requested):
Nice Work no major callouts!
One thing that would be nice to have is just a unit or integration test that covers some of the edge cases we chatted about in the design for this.

And a test to check that when metadata expires and we change the object and a new stream is opened it doesn't serve the same bytes as before.

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/MetadataStore.java
```diff
@@ -70,14 +70,11 @@ public MetadataStore(
...
-        Collections.synchronizedMap(
-            new LinkedHashMap<S3URI, CompletableFuture<ObjectMetadata>>() {
-              @Override
-              protected boolean removeEldestEntry(
-                  final Map.Entry<S3URI, CompletableFuture<ObjectMetadata>> eldest) {
-                return this.size() > configuration.getMetadataStoreCapacity();
-              }
-            });
+        Caffeine.newBuilder()
+            .expireAfterWrite(
```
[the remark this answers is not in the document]
vaibhav5140: I think we should not evict blob linked to it because we are evicting metadata based on ttl so it does not ensure that metadata has actually expired, in case S3 object is not changed metadata will be same, thus we can reuse that blob which is already in cache. Its fine to rely on memory manager for Blob cache cleanup

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/MetadataStore.java
```diff
@@ -120,9 +118,9 @@ public boolean evictKey(S3URI s3URI) {
    * @param openStreamInformation contains the open stream information
    * @return returns the {@link CompletableFuture} that holds object's metadata.
    */
-  public synchronized CompletableFuture<ObjectMetadata> asyncGet(
+  public CompletableFuture<ObjectMetadata> asyncGet(
```
stubz151, line 121: we removing the synchronized here and are letting caffeine handle the synchronization instead. Is this safe?

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/MetadataTTLTest.java
```diff
@@ -0,0 +1,179 @@
...
+import software.amazon.s3.analyticsaccelerator.S3SeekableInputStream;
+import software.amazon.s3.analyticsaccelerator.S3SeekableInputStreamConfiguration;
+import software.amazon.s3.analyticsaccelerator.common.ConnectorConfiguration;
+import software.amazon.s3.analyticsaccelerator.util.OpenStreamInformation;
+import software.amazon.s3.analyticsaccelerator.util.S3URI;
+
+public class MetadataTTLTest extends IntegrationTestBase {
+
+  @ParameterizedTest
+  @MethodSource("clientKinds")
```
stubz151, line 37: I think we use argumentsFor here so u don't have to write the same method 3 times

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/MetadataTTLTest.java
```diff
@@ -0,0 +1,179 @@
...
+        // Large objects: should throw 412 when accessing uncached blocks
+        IOException ex =
+            assertThrows(
+                IOException.class,
+                () -> {
+                  byte[] data2 = new byte[1000];
+                  int bytesRead2 = stream1.read(data2);
+                  assertTrue(bytesRead2 > 0, "Should read data before exception");
+                });
+        Throwable cause = ex.getCause();
```
stubz151, line 113: You can right this a bit nicer like this:
```
S3Exception s3Exception =
          assertInstanceOf(S3Exception.class, ex.getCause(), "Cause should be S3Exception");
assertEquals(412, s3Exception.statusCode(), "Expected Precondition Failed (412) status code");

File: input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/MetadataTTLTest.java [hunk 5]
```diff
@@ -0,0 +1,179 @@
...
+        assertTrue(cause instanceof S3Exception, "Cause should be S3Exception");
+        S3Exception s3Exception = (S3Exception) cause;
+        assertEquals(412, s3Exception.statusCode());
+      } else {
+        byte[] data2 = new byte[1000];
+        int bytesRead = stream1.read(data2);
+        assertTrue(bytesRead > 0, "Should read from prefetched cache without 412 error");
```
stubz151, line 120:
```
assertDoesNotThrow(() -> stream1.read(data2);)

Think this is slightly nicer

File: input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/MetadataTTLTest.java [hunk 6]
```diff
@@ -0,0 +1,179 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package software.amazon.s3.analyticsaccelerator.access;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import java.io.IOException;
+import java.util.*;
```
stubz151, line 21: check * imports here

input-stream/src/test/java/software/amazon/s3/analyticsaccelerator/io/physical/data/MetadataStoreTest.java
```diff
@@ -151,4 +152,111 @@ public void testHeadRequestCallbackCalled() throws IOException {
...
+    // Get new version of object
+    ObjectMetadata second = metadataStore.get(key, OpenStreamInformation.DEFAULT);
+    assertEquals("new-etag", second.getEtag());
+    assertEquals(200, second.getContentLength());
+
+    verify(objectClient, times(2)).headObject(any(), any());
+  }
+
+  @Test
+  public void testZeroTtlDisablesMetadataCache() throws IOException {
```
stubz151, line 242: It would be nice to make sure each stream is pinned to a separate etag aswell. Right now this test makes sure we made 3 head requests but can we also verify that each stream makes use of these results?

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/MetadataStore.java
vaibhav5140: earlier we were using LinkedHashMap which required manual synchronized for thread safety, now we use Caffeine which is thread-safe internally so i think we don't need synchronize

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/MetadataTTLTest.java
vaibhav5140: argumentsFor is for cartesian products of 4 parameters, but these tests only need 2 (clientKind and s3Object). We end up passing dummy StreamReadPatternKind and AALInputStreamConfigurationKind values that aren't used. Plus we still need the map lookup for the boolean logic 4MB=false, 16MB=true, so the cartesian product doesn't really help. Looking at ReadCorrectnessTest file we use separate methods , it's cleaner and avoids the unused parameters

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/MetadataTTLTest.java
vaibhav5140: Addressed in latest commit

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/MetadataStore.java
```diff
@@ -109,7 +106,8 @@ public ObjectMetadata get(S3URI s3URI, OpenStreamInformation openStreamInformati
    * @return a boolean stating if the object existed or not
    */
   public boolean evictKey(S3URI s3URI) {
-    return this.cache.remove(s3URI) != null;
+    CompletableFuture<ObjectMetadata> removed = cache.asMap().remove(s3URI);
```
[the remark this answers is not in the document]
vaibhav5140: asMap().remove() returns the removed value (null if key was not present), while invalidate() returns void. Since evictKey() needs to return a boolean indicating if the key existed, we need asMap().remove() to check the return value. invalidate() can not tell us if the key was actually there

vaibhav5140:
> Hi @vaibhav5140 I think this is a good PR overall, i am happy to approve once we have more test coverage at the stream level.
Thanks @fuatbasik for reviewing, I have moved the stream level tests to the appropriate test files, ensuring proper separation of concerns between metadata store and stream level testing. The MetadataStoreIntegrationTest now focuses only on metadata store-specific behaviour, while stream level semantics are tested in the appropriate stream test files

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/EtagChangeTest.java
```diff
@@ -181,4 +195,140 @@ protected void testChangingEtagAfterStreamPassesAndReturnsCachedObject(
       assertChecksums(checksum, cachedChecksum);
     }
   }
+
+  static Stream<Arguments> streamTtlTests() {
+    List<Arguments> testCases = new ArrayList<>();
+
+    for (S3ClientKind clientKind : getS3ClientKinds()) {
```
fuatbasik, line 202: i think the proper way would be have arguments for 4MB, False -- 4MB, True -- 4MB, null. 16MB False -- 16MB True -- 16MB Null etc.

fuatbasik: seems like your shouldThrow412 is actually not a test parameter but actually how we understand what to expect from the result. Cant we confirm that from object size?

fuatbasik (commented):
Hi @vaibhav5140 . I put some comments.

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/StreamConsistencyTest.java
```diff
@@ -0,0 +1,323 @@
...
+  }
+
+  /**
+   * Test cross-stream consistency when object changes after metadata expiry
+   *
+   * @param s3ClientKind S3 client kind to use
+   * @param s3Object S3 object to read
+   * @param ttlTime TTL time in milliseconds
+   */
+  protected void testNewStreamDetectsObjectUpdateAfterMetadataExpiry(
```
ahmarsuhail, line 223: can you add a HEAD request assertion here

ahmarsuhail (changes requested):
some minor comments

ahmarsuhail:
oh and you should probably add some documentation too. can do it as a follow up maybe

stubz151 (approved):
LGTM TY

input-stream/src/integrationTest/java/software/amazon/s3/analyticsaccelerator/access/StreamConsistencyTest.java
```diff
@@ -0,0 +1,323 @@
...
+
+  /**
+   * Test single stream consistency when metadata expires during stream lifetime
+   *
+   * @param s3ClientKind S3 client kind to use
+   * @param s3Object S3 object to read
+   * @param shouldThrow412 whether the test should expect a 412
+   * @param ttlTime TTL time in milliseconds
+   */
+  protected void testStreamBehaviorAfterMetadataExpiry(
```
ahmarsuhail, line 73: you should update this test to better reflect the S3A case we just talked about. at the  end, after TTL expiry, open a new stream stream2, and use it to read. the read should be successful.

ahmarsuhail: ah ok, sounds good

ahmarsuhail (changes requested):
looks good, could you just update your concurrency test. and then will approve

ahmarsuhail (approved):
+1, LGTM

PR #348, opened 2025-08-15
Introduce max read bytes config for sequential read

ozkoca:
## Description of change
This PR implements a maximum generation limit for sequential read progression to prevent unbounded growth of prefetch sizes while maintaining optimal performance.

### DIFF
--- a/doc/CONFIGURATION.md
+++ b/doc/CONFIGURATION.md
@@ -70,6 +70,7 @@ Options under `<CONNECTOR_PREFIX>.physicalio.`
 | `partsizebytes`                     | `8MB`   | Size of individual parts for transfer                                              |
 | `sequentialprefetch.base`           | `2.0`   | Base factor for sequential prefetch sizing                                         |
 | `sequentialprefetch.speed`          | `1.0`   | Speed factor for sequential prefetch growth                                        |
+| `sequentialprefetch.max.size`       | `128MB` | Maximum size (in bytes) for sequential prefetching                                 |
 | `max.memory.limit`                  | `2GB`   | Maximum memory limit (in bytes) that BlobStore can utilize for storage             |
 | `cache.timeout`                     | `1s`    | Time duration (in milliseconds) a block remains in BlobStore after its last access |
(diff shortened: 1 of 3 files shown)


common/src/main/java/software/amazon/s3/analyticsaccelerator/common/telemetry/TelemetryDatapoint.java
```diff
@@ -117,7 +117,9 @@ public T build() {
       return buildCore();
     }
 
-    /** @return new instance of whatever this builder builds */
+    /**
```
vaibhav5140, line 120: you need to build it with java-8

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/PhysicalIOConfiguration.java
```diff
@@ -121,6 +122,11 @@ public class PhysicalIOConfiguration {
 
   private static final String SEQUENTIAL_PREFETCH_SPEED_KEY = "sequentialprefetch.speed";
 
+  /** Defines the maximum size (in bytes) for sequential prefetching */
+  @Builder.Default private long sequentialPrefetchMaxSize = DEFAULT_SEQUENTIAL_PREFETCH_MAX_SIZE;
+
+  private static final String SEQUENTIAL_PREFETCH_MAX_SIZE_KEY = "sequentialprefetch.max.size";
```
vaibhav5140, line 128: Add this new config parameter in [CONFIGURATION.md](https://github.com/awslabs/analytics-accelerator-s3/blob/main/doc/CONFIGURATION.md)

input-stream/src/main/java/software/amazon/s3/analyticsaccelerator/io/physical/data/BlockManager.java
vaibhav5140, line 1: Can we add unit tests for block manager to verify the new maxGeneration functionality? Specifically test that sequential reads respect the sequentialPrefetchMaxSize limit and that request sizes are properly capped

PR #367, opened 2026-03-05
Standardize exception for negative seek positions to EOFException

### LINKED ISSUE #84: `S3SeekableInputStream.seek` exception discrepancy for negative positions.
S3A throws [`EOFException`](https://github.com/apache/hadoop/blob/f6c45e0bcf4aeaba31515e548dcc98b33245fe0e/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AInputStream.java#L330), whereas S3FIleIO throws [`IllegalArgumentException`](https://github.com/apache/iceberg/blob/main/aws/src/main/java/org/apache/iceberg/aws/s3/S3InputStream.java#L83).

We currently throw `EOFException`.

pratham76:
## Description of change

Changed `SeekableInputStream` and `S3SeekableInputStream` to throw `EOFException` instead of `IllegalArgumentException` when seeking to negative positions. This matches S3A behavior.

#### Relevant issues
Fixes #84

No

This matches S3A behavior and provides clearer semantics. Code catching `IllegalArgumentException` will need to catch `EOFException` instead.

No - only changes exception type for existing error case.

- Added 4 new unit tests in `SeekNegativePositionTest`

---


pratham76:
@stubz151 Could you please review? Thanks!

pratham76:
@stubz151 Gentle Ping!

ahmarsuhail:
Hey @pratham76, thank you for your contribution! Could you clarify why we need to make this change? I think the reason we couldn't decide on what to throw is because different stream implementations have different expecations. 

S3FileIO throws a `IllegalArgumentException`, see https://github.com/apache/iceberg/blob/32f5ee6c2d8d8d0476795efc4b1420e6c7f0b1c0/aws/src/main/java/org/apache/iceberg/aws/s3/S3InputStream.java#L113

